PR #17, opened 2024-03-21
Fire base setup sp1

tosigame:
Mostly settings on the firebase console (WanderPals project), however added dependency/ google services file on the project.
Also did a quick test (locally) and seems like everything was set up correctly.
dependecies where added with new toml file syntax (which was used when creating the project)


mtv47:
From Sonar: "The resource `R.string.web_client_id` appears to be unused" is normal since the sign in will be done in another pr but the "Group dependencies by their destination." is easily fixable and should be done

mtv47:
All the "implementation" should just be grouped together in the dependencies

PR #19, opened 2024-03-21
Implement google authentication

Polo150202:
I have implemented the google sign in with two screens, one with the sign in button and the other with a greeting screen.


app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -31,6 +87,34 @@ fun Greeting(name: String, modifier: Modifier = Modifier) {
   Text(text = "Hello $name!", modifier = modifier)
 }
 
+@Composable
+fun Google_button(onClick: () -> Unit) {
```
Exino329, line 91: Need Documentation for Google_button (there should be a documentation convention concerning Kotlin and composable)

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -31,6 +87,34 @@ fun Greeting(name: String, modifier: Modifier = Modifier) {
   Text(text = "Hello $name!", modifier = modifier)
 }
 
+@Composable
```
Exino329, line 90: There should be documentation for Google_button function. There is probably documentation convention concerning Kotlin/ Composable that can be followed. Should also use CamelCase naming instead of snake case

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -31,6 +87,34 @@ fun Greeting(name: String, modifier: Modifier = Modifier) {
...
+fun Google_button(onClick: () -> Unit) {
+  Row(
+      horizontalArrangement = Arrangement.spacedBy(0.dp, Alignment.CenterHorizontally),
```
Exino329, line 93: spacedBy a padding of 0.dp is useless, Use only the horizonzal arrangement.

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -1,25 +1,81 @@
...
+
+    signInClient = GoogleSignIn.getClient(this, gso)
+
     setContent {
       WanderPalsTheme {
         // A surface container using the 'background' color from the theme
         Surface(modifier = Modifier.fillMaxSize(), color = MaterialTheme.colorScheme.background) {
-          Greeting("Android")
+          navController = rememberNavController()
+          NavHost(navController = navController, startDestination = "sign_in") {
```
Exino329, line 73: "sign_in" should also be in CamelCase and be a constant value. However not being a val is for the moment fine since

Exino329 (changes requested):
There is no tests, however they will be added soon when the structure of the project will be done

Exino329 (approved):

Polo150202:
I take care of your requirements, let me know if something is wrong

PR #54, opened 2024-03-26
Data classes

tosigame:
Added data classes, that will be used by the application, and the its database, also addeds comments explaining the classes. If modifications are suggested it is easy to add new fields/ remove old ones. If needed to refactor in the future, it should also be doable (without too much effort).


Exino329 (approved):
Everything is clean and commented. Obvisously some change will probably needed in the future to accomodate new features of the project

PR #58, opened 2024-03-27
Navigation Classes and Trip Screen Navigation

mtv47:
What was done during this commit:
* Moved SignIn into its own class for cleaner MainActivity for navigating
* Created all necessary classes for Navigating inside of the app.
* Made a Trip Screen that will be attached to a selected Trip in the Overview Later.
* Trip Screen has a functional navigation bar to navigate to other related screens to be implemented later that are related to the specific trip
* Tests for the Trip Screen

What is missing and will be added further down:
* Back button to return to the Overview but Overview is missing 
* Navation to the Trip Screen after selecting a Trip or creating one


mtv47:
Further tests will be added in the futur to increase coverage when the Trip Screen and Overview are connected together to the Main and between themselves. Sonar says "Remove this unused function parameter "oldNavActions". " but oldNavActions will be used further down to go back to the Overview Screen

app/src/main/java/com/github/se/wanderpals/ui/navigation/Destination.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Data class representing a destination in the app.
+ *
+ * @param route The route of the destination.
+ * @param icon The icon of the destination.
+ * @param text The text of the destination.
+ */
+data class Destination(val route: String, val icon: ImageVector, val text: String)
+
+/** List of top level destinations in the app. */
+val TOP_LEVEL_DESTINATIONS =
```
Polo150202, line 21: You should modify the icon parameter because it displays the same thing, maybe find pics on internet to represent the different icon

app/src/main/java/com/github/se/wanderpals/ui/navigation/Destination.kt
Polo150202, line 1: Maybe you can change from kotlin file to kotlin dataclass

app/src/androidTest/java/com/github/se/wanderpals/screens/TripScreen.kt
Polo150202, line 1: You can add documentations on the tests you provide to explain the purpose of the test class and change the name to TripScreenTest

app/src/main/java/com/github/se/wanderpals/ui/navigation/Route.kt
```diff
@@ -1,3 +1,13 @@
 package com.github.se.wanderpals.ui.navigation
 
-object Route {}
+/** Object defining the routes in the app. */
+object Route {
+  const val NOTIFICATION = "notif"
```
Polo150202, line 5: Can change to: const val NOTIFICATION = "notification" explicitly

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Trip.kt
```diff
@@ -0,0 +1,66 @@
...
+@Composable
+fun Trip(oldNavActions: NavigationActions) {
+  val navController = rememberNavController()
+  val navActions = NavigationActions(navController)
+
+  Scaffold(
+      modifier = Modifier.testTag("tripScreen"),
+      topBar = {},
+      bottomBar = { BottomBar(navActions) },
+      floatingActionButton = {}) { innerPadding ->
```
Polo150202, line 34: floatingActionButton should be removed if not used later

Polo150202 (changes requested):
Just minor changes to improve the code cleanliness

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Trip.kt
mtv47: Yes, will do

app/src/main/java/com/github/se/wanderpals/ui/navigation/Destination.kt
mtv47: I'm going to remove it to make the code more compact for now then, since it's not currently used and I don't see any potential use from it

app/src/main/java/com/github/se/wanderpals/ui/navigation/Destination.kt
mtv47: Unfortunately I can't because the TRIP_DESTINATIONS is in the same file so it isn't detected as a Kotlin data class file while still being a data class Destination, with this additional list that from my point of view wouldn't make sense to put elsewhere. If you want me to still move TRIP_DESTINATIONS to accomplish the requested change, please indicate where to put the list in question

app/src/androidTest/java/com/github/se/wanderpals/screens/TripScreen.kt
mtv47: I find that documenting the tests isn't necessary because the name of the tests are expressive enough and long by design. Please tell me if you would still want me to write documentation, make the tests function names even more expressive or if nothing should be changed

mtv47: After talking to others, I would prefer to keep using TripNavigationTest as a name because only navigation is tested with very basic display checks. Further tests can be done in a future TripScreenTest or other files to test more in depth functionality not currently present

Polo150202 (approved):
All good for me

PR #59, opened 2024-03-28
Create trip repository and associated data classes

tosigame:
This PR introduces a comprehensive integration and testing setup for managing trip information within a Firebase Firestore in an Android application. The core functionality revolves around the TripsRepository class, which provides asynchronous methods to add, update, delete, and retrieve trips by utilizing Kotlin coroutines for efficient network communication. A notable addition is the FirestoreTrip DTO class, designed to bridge the gap between the app's data model and Firestore's serialization requirements, particularly converting LocalDate fields to String for compatibility. The PR also includes unit tests within TripsRepositoryTest, ensuring the repository's operations perform as expected, including trip addition, retrieval, update, and deletion functionalities, all while demonstrating the effective use of Robolectric for testing Android-specific code without the need for an emulator or physical device. (thanks to mtv47 for helping,and explaining on how to test firebase with unit tests)


app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
Exino329, line 1: When trying to request data from FireStore (i.e. inside the  try/catch blocks) It would be useful to add the Logs in case of success and failure to using `Log(...)` to track the Firebase activity with Logcat.

app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -0,0 +1,262 @@
...
+   * @return Boolean indicating success or failure of the operation.
+   */
+  suspend fun addTrip(trip: Trip): Boolean =
+      withContext(Dispatchers.IO) {
+        try {
+          // Generate a unique ID for the trip
+          var uniqueID = UUID.randomUUID().toString()
+
+          var existingDocument = tripsCollection.document(uniqueID).get().await()
+          while (existingDocument
```
Exino329, line 103: This is maybe a bit overkill since UUID guarantess uniqueness of IDS. The probability to get a colision probably shouldn't being considered according to this project

Exino329 (changes requested):
Very Good code quality overall ! However there is also some additional things that must be done/considered :

1. Reducing code smells from Sonarcloud.
2. Add some tests to Firestore DataClasses.
3. Potentially check for renaming the methods `to..` and  `from` from the FirestoreData classes since it introduces some redundancies when calling (e.g. FirestoreTrip.toTrip(trip)).

tosigame:
The get method in Firestore transactions is specifically designed for retrieving documents within a transaction, ensuring transactional integrity, and does not align with the semantics of indexed accessors used in collections. Consequently, we're ignoring the suggestions to replace function calls with indexed accessors as two identified code smells, as they don't apply in this context.

Exino329 (approved):
Can merge with main !

PR #60, opened 2024-03-29
Creating trip creation screen with inputs

mtv47:
I've created a simple Trip Creation Screen with the associated view model linked to the TripRepository for firebase access. 

It's missing the functionality of inviting friends because there is some challenges with the fact that the databases stores for now only ids that are hard to get for other users when you are a user. Inviting users by giving them the trip id to enter in the overview is one possible solution that is being discussed but if the team decides that the functionality of inviting should be present in this screen, I will add it in a futur release

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/CreateTripTest.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/CreateTripTest.kt
@@ -1,0 +1,167 @@
+package com.github.se.wanderpals
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.se.wanderpals.model.data.Trip
+import com.github.se.wanderpals.model.repository.TripsRepository
+import com.github.se.wanderpals.model.viewmodel.CreateTripViewModel
+import com.github.se.wanderpals.screens.CreateTripoScreen
+import com.github.se.wanderpals.ui.navigation.NavigationActions
+import com.github.se.wanderpals.ui.navigation.Route
+import com.github.se.wanderpals.ui.screens.CreateTrip
+import com.kaspersky.components.composesupport.config.withComposeSupport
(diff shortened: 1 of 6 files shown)


tosigame (changes requested):
Hey, I reviewed your code and it mostly looks good. However, there are some cases where the app can break due to user input. Therefore, you should sanitize the user input data. Also, in the event of incorrect inputs, it would be beneficial to show more information to the users. Additionally, improving and adding some code tests that could fail on bad user input is advisable. Also some refactoring could make the code cleaner and easier to read.

Tests:
1: Add test cases for scenarios when the description is missing to ensure comprehensive test coverage.
2: In CreateTripViewModelTest, check for attribute-level equality rather than object-level to minimize false positives.

CreateTripViewModel:
3: Consider merging CreateTripViewModel with OverviewModel in future releases to reduce ViewModel redundancy.

CreateTrip Screen:
4: Mark the budget field as optional to align with the flexibility in trip planning.
5: Consider making the description field optional, offering users more flexibility.
6: Implement validation to ensure the startDate is before the endDate to prevent planning errors.
7: Ensure error messages are clear and specific to guide users in correcting their input effectively.
8:Validate that the budget, if entered, is a positive number to ensure data integrity. (and a double, not text)
9: Integrate a mini-calendar beside the date input fields in a future release, allowing users a choice in date selection method.
10: Refactor code to separate the Scaffold structure from content logic, improving code maintainability.
11: Move hardcoded UI strings into strings.xml for better localization support and easier maintenance.

mtv47:
For the point 2,  In Kotlin, the == operator is used to compare the equality of two objects in terms of their structural content, not their reference identity

mtv47:
For all the elements that will be addressed later, this was chosen to have a version available for others faster to create/see trips from the emulator to facilitate live testing for PRs that require it

tosigame (approved):
You solved most of my concerns, and the reasons to not address some of them, are also valid. Great job! Also nice and smart way to handle errors !

PR #92, opened 2024-03-30
Trip repository stops methods

tosigame:
Implemented methods in the TripsRepository class to manage stop-related operations, including adding, removing, modifying, retrieving, and listing all stops for a specified trip ID. Additionally, minor corrections were made in the getAllTrips and addTrip methods. A basic test suite has been developed to validate the functionality of these new and corrected methods.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
@@ -70,7 +70,13 @@ class TripsRepository(
                   .document(stopId)
                   .get()
                   .await()
-          documentSnapshot.toObject<FirestoreStop>()?.toStop()
+          val firestoreStop = documentSnapshot.toObject<FirestoreStop>()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -58,6 +60,139 @@ class TripsRepository(
...
+                  .get()
+                  .await()
+          val firestoreStop = documentSnapshot.toObject<FirestoreStop>()
+          if (firestoreStop != null) {
+            firestoreStop.toStop()
+          } else {
+            Log.e("TripsRepository", "getStopFromTrip: Not found stop $stopId from trip $tripId.")
+            null
+          }
+        } catch (e: Exception) {
```
ForestDope, line 80: Nice use of try catch

ForestDope (approved):
Overall very good code, systematic error catching and handling is good, structure is also clear. The tests also seem fine and should be effective as every functions is tested.

PR #93, opened 2024-03-30
Overview screen development

Exino329:
The overview Screen is mainly done. The user Can see his current trip and filter them by title with a search bar. He also have 2 buttons for creating a New Trip and join a trip with a code (the tripId).  Just a note concerning this PR :
- The tests for the search bar are actually not working due to a problem of text input. But the filtering option is working on the emulator so it is acceptable for the moment. The problem concerning the textInput will be solved later.


mtv47 (changes requested):
Please look at my suggestions, overall very concise and good code. The moving of the test file however should be done. Spending more time on the tests isn't worth it at this point.

app/src/androidTest/java/com/github/se/wanderpals/trip/OverviewTest.kt
mtv47, line 1: I've also had a similar problem with my tests, it happened when the composable is complex and not created by the coder, the test doesn't understand that it's an imputable text field in your case. I've researched the problem myself before and haven't found a solution that works either. I'm going to stay in touch if I find anything

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
mtv47, line 1: MainActivity is close to the main branch one, but was branched before so not a problem and will be resolved during a merge conflict

app/src/main/java/com/github/se/wanderpals/ui/navigation/Route.kt
mtv47, line 1: Since we use Route.CREATE_TRIP in your code, it won't be a problem, I find your choice of string to be better than the current one

app/src/main/java/com/github/se/wanderpals/model/viewmodel/OverviewViewModel.kt
mtv47, line 1: TripViewModel could be merged with your OverviewViewModel, since TripViewModel didn't exist on your branch it's normal. Can be addressed in another PR when linking all the code will be done. Just to streamline the view models

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/Overview.kt
mtv47, line 1: Very nice code separation with the additional composable functions to make the Scaffold more readable, however I don't know if putting comments inside of the Scaffold is needed here, up to you

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewBottomBar.kt
mtv47, line 1: FontWeight and Color should be put inside of a stylistic file, but since the theme of the application isn't defined up to this point, it isn't in my opinion a problem. If you want you can put it as a variable like DATE_PATTERN since it's the same multiple times but it's not strictly necessary

The use of function to avoid passing too many arguments is a nice touch

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewContent.kt
mtv47, line 1: I don't know if having a letterSpacing of 0.5 is really needed but I guess it was taken from Figma

mtv47 (changes requested):

app/src/androidTest/java/com/github/se/wanderpals/trip/OverviewTest.kt
Exino329: Ok I'll also investigate in the future to solve this issue. I posted something on ed  too.

app/src/main/java/com/github/se/wanderpals/model/viewmodel/OverviewViewModel.kt
Exino329: Yeah that's a goodi idea, We will look later how to do it.

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewContent.kt
Exino329: Yes it was taken directly from figma (with Dev mode) so I think we could discuss about policies and colors later but these are details I think

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewBottomBar.kt
Exino329: Yes but since we did not defined concrete  theme for the moment I prefer to let it like this.

app/src/androidTest/java/com/github/se/wanderpals/trip/OverviewTest.kt
mtv47: I don't think putting OverviewTest inside of trip is a great idea, maybe create a another folder

Exino329: Yes effectively, I'll do it !

mtv47 (approved):
Great work

PR #99, opened 2024-03-30
Map screen dev

Polo150202:
Add the Map API, with a button to modify the display of the zoom button (Format edition).


tosigame (changes requested):
Hello,

The initial setup for integrating maps into our project is well-constructed. However, we've encountered an issue due to the inclusion of certain files in the repository that are typically excluded via .gitignore. Specifically, these files contain project settings that reference incorrect paths to the Android SDK, leading to a situation where the project fails to compile and run.

To address this problem, we suggest the following actions:
1) Remove local.properties: Delete this file from the repository and ensure it's listed in .gitignore to avoid sharing environment-specific settings.

2) Handle Maps API Key Locally: Each developer should manually insert the Maps API secret key into their local workspace, keeping it out of version control for security.

Or 

The app is running smoothly after I updated the sdk.dir in the local.properties file to the correct SDK path. It's worth noting that if no SDK path is specified, the system defaults to using the user's default SDK location. Therefore, you might consider simply removing the sdk.dir= line from the local.properties file. This way, we might not need to require every team member to manually copy and configure the SDK path individually.

tosigame (approved):

tosigame:
Tested on my end it worked, We will just need to inform all team members to sync graddle, after pulling from Main, good job!

PR #102, opened 2024-04-01
Agenda screen : calendar widget

ForestDope:
Calendar widget is implemented and tested. User can navigate through months and select a specific date.


app/src/main/java/com/github/se/wanderpals/model/viewmodel/AgendaViewModel.kt
Exino329, line 1: Documentation of the class and functions is missing in this file.

Exino329 (changes requested):
Very good code overall. The two main important things to resolve are : 
- Modularize the Agenda.kt file
- Documenting the agendaViewModel

Exino329 (approved):

PR #103, opened 2024-04-01
stopped tracking local properties

tosigame:
Removed tracking of local.properties, because people starting pushing it, and so we remove it, and if someone didn't get it, we will just have them add it locally


tosigame:
we can't actually delete the file, because of map key so we will have everyone run
git update-index --assume-unchanged local.properties
to ignore local changes

PR #105, opened 2024-04-03
Fix trips repository add trip id method

tosigame:
Fixed AddTripId method, for ensuring that only valid trip IDs are added, simplified the base tests for the Tripid methods, the normal functionnality is tested throught the other methods, because they interdepend heavily on each other. More tests cases will be added in the future.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
@@ -564,62 +564,62 @@ class TripsRepository(
         Log.d("TripsRepository", "addTripId: Adding tripId to user")
 
         if (!isTripIdValid(tripId)) {
-          false
-        } else {
+          Log.d("TripsRepository", "addTripId: isTripIdValid returned false")
+          return@withContext false
+        }
 
-          val userDocumentRef = usersCollection.document(uid)
+        val userDocumentRef = usersCollection.document(uid)
 
-          // Ensure the user's document exists before attempting to modify it.
-          // If the document does not exist, create it with an initial empty list of tripIds.
-          val userDoc = userDocumentRef.get().await()
-          if (!userDoc.exists()) {
-            // Initialize the document with an empty tripIds list.
-            userDocumentRef.set(mapOf("tripIds" to listOf<String>())).await()
-          }
+        // Ensure the user's document exists before attempting to modify it.
+        // If the document does not exist, create it with an initial empty list of tripIds.
+        val userDoc = userDocumentRef.get().await()
+        if (!userDoc.exists()) {
+          // Initialize the document with an empty tripIds list.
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -522,6 +522,33 @@ class TripsRepository(
...
+   * Checks if the specified trip ID exists in the 'Trips' collection. This method queries the
+   * Firestore database to verify the existence of a document corresponding to the given trip ID. It
+   * ensures that operations related to trip IDs are conducted with valid identifiers.
+   *
+   * @param tripId The unique identifier of the trip to be validated.
+   * @return Boolean indicating whether the trip ID is valid (true) or not (false). If the trip ID
+   *   exists in the database, returns true; otherwise, returns false. In the event of an exception
+   *   during the database query, the method also returns false,
+   */
+  private suspend fun isTripIdValid(tripId: String): Boolean =
```
mtv47, line 535: I think that making the function public won't expose anything significant and could be helpful in the future to reduce unnecessary PRs to open it up. It is, however, at your discretion

app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -522,6 +522,33 @@ class TripsRepository(
...
+      withContext(dispatcher) {
+        try {
+          val document = tripsCollection.document(tripId).get().await()
+          if (document.exists()) {
+            Log.d("TripsRepository", "isTripIdValid: tripId $tripId exists.")
```
mtv47, line 540: Very nice Logging, greatly appreciated, having the logging done before the returns in the functions that use it could be interesting to keep up the pattern that I see with the rest of the code for clarity sake and more in-depth direct messages. But really not mandatory

app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -536,58 +563,63 @@ class TripsRepository(
...
+            // Initialize the document with an empty tripIds list.
+            userDocumentRef.set(mapOf("tripIds" to listOf<String>())).await()
+          }
 
-                    val rawTripIds = snapshot["tripIds"]
-                    if (rawTripIds is List<*>) {
-                      // Filter non-null and String values only, safely adding them to
-                      // existingTripIds.
-                      existingTripIds.addAll(rawTripIds.filterIsInstance<String>())
+          try {
```
mtv47, line 580: Cognitive Complexity is seen as a problem with Sonar but the fact that your code has great variable names and great structure described with comments renders this issues not needing a fix from my end

mtv47 (dismissed):
No changes needed for me explicitly, however I have taken the liberty to make some suggestions

mtv47 (approved):
Great job!

PR #108, opened 2024-04-05
added User Like list to suggestion and updated relevant tests

tosigame:
Added like field to suggestions, as decided on the team meeting


Exino329 (approved):
Just some updates concerning the data class with Firestore. I have nothing special to comment concerning these updates.

PR #111, opened 2024-04-07
Clean + upgrade for the Create Trip and Overview Screen 

Exino329:
In this PR, the following updates have been made :

1. The description not single lined anymore (it was a small visual bug fix) And the size of the title is now limited to 35 caracters).
2. The overview screen is upgraded : Now there is a button for each trip at the right allowing someone to share the Trip to friends (using the trip id) via email or other app (such as whats app for example).
3. The CreateTripViewModel has been merged in the OverviewViewModel.
4. Padding was adjusted in the OverviewViewModel to make the display cleaner when scrollable column showing the trips is full.


app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTrip.kt
```diff
@@ -68,21 +98,37 @@ fun OverviewTrip(trip: Trip, navigationActions: NavigationActions) {
...
+                              fontSize = 12.sp,
+                              lineHeight = 24.sp,
+                              fontWeight = FontWeight(500),
+                              color = Color(0xFF000000),
+                              letterSpacing = 0.5.sp,
+                          ))
+                  // Share trip code button
+                  IconButton(
+                      modifier = Modifier.size(20.dp).testTag("shareTripButton" + trip.tripId),
+                      onClick = { context.shareTripCodeIntent(trip.tripId) },
```
annhit29, line 122: It's good, one thing that can be improved is: there's a background color (let's say grey) for the share icon when clicking on it so that the user will not be confused about which thing (the box button for going to the dashboard or the icon button for sharing the trip) he/she is clicking

app/src/main/java/com/github/se/wanderpals/ui/screens/CreateTrip.kt
```diff
@@ -123,13 +128,23 @@ fun CreateTrip(tripViewModel: CreateTripViewModel, nav: NavigationActions) {
             horizontalAlignment = Alignment.CenterHorizontally) {
               OutlinedTextField(
                   value = name,
-                  onValueChange = { name = it },
+                  onValueChange = { newTitle ->
+                    if (newTitle.length <= MAX_TITLE_LENGTH) {
+                      name = newTitle
+                    }
```
annhit29, line 134: can create an else case for the case when newTitle.length > MAX_TITLE_LENGTH

annhit29 (changes requested):
All tests have passed on my side.
Two things can be improved: 
1. for your first point: when the title size exceeds 35 characters, there can be a red text displayed to say "at most 35 characters" to be clear (also see the second review message)
2. for your second point: see the first review message

The coverage is 76.1%, almost 80%!

Good job :D

app/src/main/java/com/github/se/wanderpals/ui/screens/CreateTrip.kt
Exino329: Here an else is not necessary since we just want to limit the maximum possible size, (i.e. if the title length is higher than 35, don't update)

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTrip.kt
Exino329: Nice catch ! I did not thought about this small detail.

app/src/main/java/com/github/se/wanderpals/ui/screens/CreateTrip.kt
Exino329: Also note that there is no need to advertise the user that he has reached the maximum length for the title. Since there is already a little text showing the title length as a fraction

annhit29 (approved):
I tried and tested on my emulator, everything is good now!
And for you final version (just before merging), the line coverage is 90.4% :D

app/src/main/java/com/github/se/wanderpals/ui/screens/CreateTrip.kt
annhit29: Yes, I didn't see the character count `"${name.length}/$MAX_TITLE_LENGTH"` with `MAX_TITLE_LENGTH = 35`, because it was a bit low i.e. it was much closer to the Budger field. We (one of the team members) can move it a bit higher to the Title field later :D

PR #112, opened 2024-04-07
Fixed Authentification to corretly use FirebaseAuth

tosigame:
Login in with google now correctly authentificates the app to use the read/write operations, using the context of the authentificated user. also removed auto sign out.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -178,6 +178,11 @@ dependencies {
     // Coroutine support for Firebase (Play Services)
     implementation(libs.kotlinx.coroutines.play.services)
 
+    // Firebase Authentication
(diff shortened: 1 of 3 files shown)


app/build.gradle.kts
annhit29, line 1: Good

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
annhit29, line 1: Good

gradle/libs.versions.toml
annhit29, line 1: Good too

annhit29 (approved):
The code works correctly on my side. It seems all good.

PR #118, opened 2024-04-10
Map with places

mtv47:
Replaced the OpenStreetMap that worked with https://nominatim.org by Google Places


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Map.kt
```diff
@@ -85,15 +91,22 @@ fun Map(tripId: String, mapViewModel: MapViewModel) {
...
     }
-    requestLocation(
-        searchText,
-        onSuccessRequest = { output, loc ->
-          // take all the suggestions
-          listOfPropositions = output
+    getAddressPredictions(
+        client,
+        inputString = searchText,
+        location = LatLng(46.519653, 6.632273),
```
Polo150202, line 97: It should be a variable, even though it would be later replaced by the position of the city where the trip is originally planned or the coordinate of the map view

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Map.kt
```diff
@@ -184,53 +212,33 @@ fun Map(tripId: String, mapViewModel: MapViewModel) {
...
+      val request =
+          FindAutocompletePredictionsRequest.builder()
+              .setLocationBias(
+                  location?.let { locationBias ->
+                    RectangularBounds.newInstance(
+                        LatLng(locationBias.latitude - 0.1, locationBias.longitude - 0.1),
+                        LatLng(locationBias.latitude + 0.1, locationBias.longitude + 0.1))
+                  })
+              .setOrigin(location)
+              .setCountries("CH")
```
Polo150202, line 234: As discussed, it would be great to add the countries covered by the trips

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Trip.kt
```diff
@@ -26,10 +26,16 @@ import com.github.se.wanderpals.ui.navigation.NavigationActions
 import com.github.se.wanderpals.ui.navigation.Route
 import com.github.se.wanderpals.ui.navigation.TRIP_DESTINATIONS
 import com.github.se.wanderpals.ui.screens.trip.agenda.Agenda
+import com.google.android.libraries.places.api.net.PlacesClient
 
 /** The Trip screen. */
 @Composable
-fun Trip(oldNavActions: NavigationActions, tripId: String, tripsRepository: TripsRepository) {
+fun Trip(
```
Polo150202, line 33: great idea to put it in the Trip file, then others can use it if needed

Polo150202 (approved):
Globally good

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Map.kt
mtv47: Will try to replace it with the camera's coordinates

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Map.kt
```diff
@@ -143,8 +156,22 @@ fun Map(tripId: String, mapViewModel: MapViewModel) {
             // suggestions
             if (expanded) {
               listOfPropositions.forEach {
-                // create a button
-                Column { Button(onClick = { searchText = it }) { Text(text = it) } }
+                val primaryText = it.getPrimaryText(null).toString()
+                val placeId = it.placeId
+                val placeFields = listOf(Place.Field.LAT_LNG)
```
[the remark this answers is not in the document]
mtv47: We can work on it on your branch directly when we meet on campus

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Map.kt
mtv47: I have limited it to CH to reduce the cost of the query for now, of course it will become more precise later as needed

PR #121, opened 2024-04-11
Dashboard suggestion widget

AgeX21:
In this pull request the following things have been done :

- The dashboard now displays up to the last 3 suggestions in a compacted form with a minimal amount of data 

- The widget is clickable and redirects to the suggestion screen when clicked

- Some graphical improvement on the suggestion popup UI, according to what was discussed in the meeting


app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestion.kt
```diff
@@ -0,0 +1,86 @@
...
+@Composable
+fun DashboardSuggestion(suggestion: Suggestion) {
+  Card(modifier = Modifier.padding(8.dp).testTag("suggestion${(suggestion.suggestionId)}")) {
+    Column(
+        modifier =
+            Modifier.border(1.dp, Color.Black, shape = RoundedCornerShape(16.dp)).padding(16.dp)) {
+          Row(
+              modifier = Modifier.fillMaxWidth(),
+              horizontalArrangement = Arrangement.SpaceBetween) {
+                Text(text = suggestion.stop.title, modifier = Modifier.testTag("stopTitle"))
```
annhit29, line 29: You can limit the lines to only receive a certain number of characters `MAX_ONE_LINE_LENGTH` for `suggestion.stop.title`. If it reaches this number, it will go to the next line to display the remaining `suggestion.stop.title`.

See the first point of the overall comment. This will resolve the issue mentioned in it.

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionDetailPopup.kt
annhit29, line 1: Great :D, the popup is prettier.

annhit29 (changes requested):
Overall comment:

All the tests passed on my side!


1. The dashboard now displays up to the last 3 suggestions in a compacted form with a minimal amount of data
Can be improved in DashboardSuggestion.kt:
When the title `suggestion.stop.title` is too long (let's say when it is more than `val MAX_ONE_LINE_LENGTH = 18`), limit the title being able to display per line to MAX_ONE_LINE_LENGTH, so you can bring the remaining title to the second line of the widget.
As for now, the title "Umeda Sky Building 2" is glued together with the "Created at : Sep" with the remaining date not displayed on the same line. You can see Create.Trip.kt and we limited the title to at most 35 characters `val MAX_TITLE_LENGTH = 35`, so by fixing `val MAX_ONE_LINE_LENGTH` to 18, we ensure that the title will display at most on two lines on the dashboard's suggestion.
What you can do is to change the "Created at :" to "on: " to earn more space in the widget.


And because the date is pushed by the long title to the second line, the `userName` is also pushed to a lower place (eg: widget1). You can use a Spacer
It would be good to unify them for the three suggestions of the dashboard.

2. The widget is clickable and redirects to the suggestion screen when clicked
Good, but when we switch from dashboard screen to suggestion screen, it launches twice to show the suggestion screen.
Also, when we switch from suggestion screen to dashboard screen, it also launches twice to show the dashboard screen.


3. Some graphical improvement on the suggestion popup UI, according to what was discussed in the meeting
Now the "Address" and the "Website" are displayed after "Comments" and after "Scheduled from".


70.0% Coverage on the new code is good, the file that decreases the coverage is DashboardSuggestion.kt which is at 46.1%, because there is DashboardSuggestionPreview() with values defined inside. This file is has the lowest coverage and is under 50%, and it decreases very much the coverage.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Dashboard.kt
```diff
@@ -18,18 +23,51 @@ import androidx.compose.material3.MaterialTheme
...
+import com.github.se.wanderpals.model.viewmodel.DashboardViewModel
 import com.github.se.wanderpals.ui.navigation.NavigationActions
 import com.github.se.wanderpals.ui.navigation.Route
+import com.github.se.wanderpals.ui.screens.dashboard.DashboardSuggestionWidget
+import java.time.LocalDate
+import java.time.LocalTime
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
+val stop =
```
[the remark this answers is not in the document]
AgeX21: Removed

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestion.kt
```diff
@@ -0,0 +1,86 @@
...
+                Text(
+                    text =
+                        "Created at : " +
+                            suggestion.createdAt.format(
+                                DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM)),
+                    modifier = Modifier.testTag("createdAt"))
+              }
+          Row(
```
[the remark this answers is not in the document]
AgeX21: Fixed

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestion.kt
AgeX21: Fixed

annhit29 (approved):
Generally good, the tests passed on my end and the line coverage is 95.9%.

We will discuss for unifying the visual part much deeper with the entire team.

PR #124, opened 2024-04-11
78 generate an apk and test one end to end user flow on a real phone

mtv47:
I've changed the SDK version to a lower one, this was done to make the app compatible with more Phones and since we don't use any functionality that requires a higher SDK, it doesn't effect the implementation except for Theme.
For Theme the coded needed to change slightly but the functionality is still the same, Theme will be reworked afterwards in Sprint 4 when working on a global design.

While testing the app on the real Phones, I saw that the Search Bar of Overview wasn't centered and fixed it.

I've also updated the libraries since it didn't make sense to do a separate PR for this and it seemed to fit this one.

The Back Button was also disabled because we have multiple NavHosts and sometimes the backbutton is lost in the graph of composables, to mitigate this issue, disabling it for now seems to be the best option and we can work on having a single NavHost for futur Sprints.


Polo150202 (dismissed):
All great

PR #128, opened 2024-04-11
Fixed Bug with navigation

mtv47:
We have detected an issue when going to a Trip 1, going back to the list of Trips in Overview, and then to going Trip 2. The navigation brings you inside of the Trip 1 and not the Trip 2.

After multiple tries, the simplest solution was to store the tripId inside of the NavAction directly

@AgeX21 originally found the bug


Exino329 (approved):
Great fix. I think it is the simplest way to solve the problem. I also tried unsuccessful attemps to fix the problem on my side but in this way it is simple and does not require lot of change in the code.

PR #130, opened 2024-04-11
First try but really not sure - CI Action APK

mtv47:
Implementation of the CI Action related to generating the APK

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,57 @@
+name: Build App APK
+
+on:
+  push:
+    branches:
+      - main
+
+  pull_request:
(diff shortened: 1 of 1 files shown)


Polo150202 (approved):
It seems good to me

mtv47:
"./gradlew build" runs the unitTests each time it is run, which contains testing code for the firebase, which puts data online and then deletes it before being done. 

It's failing because the CI Action of Building the APK is executed at the same time as the CI Action Sonar so both run unitTests at the same time very often which means that one test may expect nothing to be present while another tests running on another Action has added data. Unstability and constant failure is the result

PR #133, opened 2024-04-12
Revert "First try but really not sure - CI Action APK" - And Fix

mtv47:
Reverts WanderPals/WanderPals#130 because it doesn't currently work, will delete "Upload APK to Release" since it's not needed.

Running the CI Actions of the Runner and APK at the same time make the tests for the repositories fail, need to be careful


Polo150202 (approved):
Great Job

PR #134, opened 2024-04-12
Removed testDebugUnitTest from grable of apk

mtv47:
I've found a fix for the APK Generation that should work every time now with an added exclusion of testDebugUnitTest that I didn't know was possible before.

Sorry for the 3 related PRs and commits that where done to achieve this


Exino329 (approved):

mtv47:
Follow up in #137

PR #136, opened 2024-04-12
Adding more Logging inside of MainActivity and making the CI Action APK run more often

mtv47:
Adding a lot more Logging to see why the APK generated by the CI Action can't go further than the SignIn Screen and made the CI Action of building the APK constantly run on new commits inside of this PR

APK generated locally with the same commands produces an APK that works perfectly fine

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,8 @@ on:
   pull_request:
     types:
       - closed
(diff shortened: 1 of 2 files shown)


Exino329 (approved):

mtv47:
Follow up in #137

PR #137, opened 2024-04-12
Removing app:testReleaseUnitTest and app:test from APK Build

mtv47:
Following the occasional fails of #134

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
 
       # Run Build Project
       - name: Build gradle project
-        run: ./gradlew build -x app:testDebugUnitTest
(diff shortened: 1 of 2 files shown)


Exino329 (approved):

mtv47:
This fixed the generation of the APK not being able to finish by not calling at all UnitTests now, but a problem still persists. The APK generated locally works perfectly with the sign in of Google, but when the APK is generated throw the CI, the sign in doesn't work. This problem I think is related to the origin of the APK coming from a server such as GitHubs making Google think it's a malicious application. 

In any case, this can be resolved by confirming on the email received from Google that the device in indeed yours:
image

This will be mitigated in a futur PR that adds another button as a secondary option for logining it

PR #140, opened 2024-04-12
Update README.md with Architecture Diagram

tosigame:
Added Architecture Diagram


mtv47 (approved):
Previous versions are visible in the Issue associated, the diagram was talked about during a video call and everything seems to have been drawn accordingly

PR #141, opened 2024-04-12
Suggestion comment

AgeX21:
In this pull request the following things have been done :

- Users can now see the list of comments and add their own which appears at the top of the list

- In a comment, the user can see who commented, when was it posted, and what the comment is about


app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
```diff
@@ -106,4 +108,24 @@ open class SuggestionsViewModel(
...
+            comments =
+                suggestion.comments +
+                    Comment(
+                        commentId = UUID.randomUUID().toString(),
+                        userId = currentLoggedInUId,
+                        userName = comment.userName,
+                        text = comment.text,
+                        createdAt = comment.createdAt))
+    viewModelScope.launch {
+      val wasUpdateSuccessful =
```
mtv47, line 124: Creating a variable isn't really needed here as I understand it, but it makes sense since toggleLikeSuggestion also uses it

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Trip.kt
```diff
@@ -64,7 +64,6 @@ fun Trip(
             Agenda(AgendaViewModel(tripId, tripsRepository))
           }
           composable(Route.SUGGESTION) {
-            Suggestion(oldNavActions, tripId, SuggestionsViewModel(tripsRepository, tripId))
```
mtv47, line 67: Good catch, weird that it's present inside of the main

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Suggestion.kt
```diff
@@ -66,7 +66,8 @@ fun Suggestion(
         bottomBar = {
           SuggestionBottomBar(
               onSuggestionClick = {
-                oldNavActions.navigateTo("${Route.CREATE_SUGGESTION}/$tripId")
+                oldNavActions.currentTrip = tripId
+                oldNavActions.navigateTo(Route.CREATE_SUGGESTION)
```
mtv47, line 70: Fixed according to the new use of NavActions, Great Job

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionDetailPopup.kt
```diff
@@ -52,190 +66,250 @@ fun SuggestionDetailPopup(
...
 
-                Spacer(modifier = Modifier.height(16.dp))
+                      Box(
+                          modifier = Modifier.fillMaxWidth(),
+                          contentAlignment = Alignment.CenterEnd) {
+                            Text(
+                                text = "Send comment",
+                                modifier =
+                                    Modifier
+                                        //            .padding(8.dp)
```
mtv47, line 263: Should be removed since not used

mtv47 (dismissed):
Small changes are required

mtv47 (approved):
All Great

PR #162, opened 2024-04-12
Sign in without googleclient

mtv47:
Created an additional Button that requests an email from the user to be able to Sign into the app without the use of Google without any checks so that the user has an always available sign in option in case the Google one doesn't work.

It works by hashing the given String without any checks


tosigame (approved):
Great Fix, works on emulator, and seems to be consistent across different runs. (Will need to be fully implemented in sprint 4)
Good Job!

PR #164, opened 2024-04-12
New NavActions linking everything

mtv47:
New NavAction works for linking Suggestion Creation from Map and Feed

Will be done in a better way inside of Sprint 4


tosigame (approved):
Works,  however needs to be cleaned up later (will be fixing next sprint)

PR #172, opened 2024-04-12
End to end test (not complete)

Exino329:
This PR is an unsuccessful attempt to finish end to end tests. Especially due to the time constraint and the fact that Trip composable contains an inner navigation. However, the tests from overview to trips is done so it is still useful to push this code on the main.


tosigame (approved):
Looks good, even tho its unfinished )

PR #173, opened 2024-04-13
Update user data class

tosigame:
Finished updating user data class, changing to the wanted fields, which a user in a trip should have. Also added an enum for a Users role ina  trip. also fixed teh functionnality of the repository to work with the updated users (created a Firestore User DTO), to correctly serailize enums and the other data.
Fixed and added tests to check the added functionnality.


ForestDope (approved):
Looks all fine to me. Good work !

PR #174, opened 2024-04-15
Trip notification data class

tosigame:
Added the tripNotification data class, its DTO object (FirestoreTripNotification), added a list of TripNotification Ids to trip object. Documented and tested the new code.


app/src/main/java/com/github/se/wanderpals/model/data/TripNotification.kt
annhit29, line 1: The fields correspond to what we discussed.

app/src/main/java/com/github/se/wanderpals/model/firestoreData/FirestoreTripNotification.kt
annhit29, line 1: Good with two methods to convert the string to LocalDateTime and vice versa.

annhit29 (approved):
Good! and the line coverage = 94.9%

PR #176, opened 2024-04-16
Add User to Trip when Joining or creating one

tosigame:
Added a private function that manages the adding of the user to a Trip, this function is called when AddTripId is called, it sets the User to be the OWNER when he creates the Trip, and its automaticcaly set as a member when he joins one (In which case the user is added to that Trip)


AgeX21 (approved):
LGTM! Code coverage is a little low, but that's to be expected, as writing automatic tests for Firebase would be problematic.

PR #177, opened 2024-04-17
Create trip notification repository methods

tosigame:
This pull request introduces a series of updates to the TripsRepository class to manage trip notifications within our application. Specifically, it adds methods for adding, retrieving, updating, and removing trip notifications from a Firestore database. Each method ensures robust error handling and logs appropriate messages for debugging purposes. The tests in testTripLifecycleWithTripNotifications validate the basic behaviors of these methods, ensuring that trip notifications can be added, retrieved, updated, and removed correctly, and that changes persist in the database as expected. These tests help guarantee the reliability and functionality of our trip notification management system within the application.


mtv47 (approved):
Your code is consistent with your other methods which makes it easier to understand and very homogeneous. Great use of Logging with Log.e and Log.d used depending if it's an error or just a message, I think using only these two methods is more than enough but Log.v can be maybe used the Logging messages become too long, but at the same time it will make it less concise. Overall, nothing to change or add, Great Job!

tosigame:
🚢

PR #179, opened 2024-04-18
Refactor suggestion detail screen

ForestDope:
I refactored the suggestion detail screen to be a full screen and not a popup anymore. Also redid the tests for the class. 
There are also some changes here and there to make everything work. For some reason the goBack function of the nav doesnt go back to the previous page. And when the keyboard is automatically dismissed after sending a comment there is a slight glitch, couldnt fix it, maybe its not a problem from the code.


AgeX21 (dismissed):
LGTM! Nothing to add, the sonar cloud coverage seems wrong, the test pass on my end and should cover some line of code but 0% on SuggestionDetail is weird. 
For the nav, it should be reworked soon so it'll be looked into it later.
For the keyboard, the glitch is minor and can be fixed later.
I'll approve so you can continue but look for a fix for the keyboard.

AgeX21 (approved):
It didn't change the coverage but nice try

Issue #154, opened 2024-04-12
Create Announcement Screen


tosigame:
Admin users, can create a new notification from the Notification Screen(create button only visible to them). Where they can write the message for the notification, and other fields specified in the dataclass.

//post:edit: An Annoucement is a type of Notification

PR #180, opened 2024-04-18
Create new trip notification screen

annhit29:
In this pull request,
1) I create the notification screen (CreateNotification.kt) for the administrator of a trip to announce with the use of CreateNotificationViewModel.
2) Both the title and the description of the notifications are mandatory to be filled, with the title size cannot exceed 55 characters.
Note: With the use of `trim()` function, we ensure that the mandatory fields cannot be filled with only whitespace characters.

This pull request is from the issue #154.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateNotificationViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateNotificationViewModel.kt
@@ -1,0 +1,30 @@
+package com.github.se.wanderpals.model.viewmodel
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.github.se.wanderpals.model.data.TripNotification
+import com.github.se.wanderpals.model.repository.TripsRepository
+import kotlinx.coroutines.launch
+
+/**
+ * ViewModel for creating notifications. This ViewModel is responsible for adding a notification to
+ * a trip.
+ */
+open class CreateNotificationViewModel(tripsRepository: TripsRepository) : ViewModel() {
+  private val _tripsRepository = tripsRepository
+
+  open fun addNotification(tripId: String, tripNotification: TripNotification): Boolean {
+    var a: Boolean = true
+    viewModelScope.launch {
+      _tripsRepository.addTripNotificationToTrip(tripId, tripNotification).also { a = it }
+    }
+    return a
+  }
+
+  /*
+  //This is for @Preview purposes only
+  fun addNotification(tripId: String, tripNotification: TripNotification) : Boolean{
+      return true
+  }
+  */
+}
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateNotificationViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateNotificationViewModel.kt
@@ -13,6 +13,13 @@ import kotlinx.coroutines.launch
 open class CreateNotificationViewModel(tripsRepository: TripsRepository) : ViewModel() {
   private val _tripsRepository = tripsRepository
 
+  /**
+   * Adds a notification by the administrator to a trip.
+   *
+   * @param tripId The ID of the trip to which the notification is to be added.
+   * @param tripNotification The notification object to be added.
+   * @return A boolean representing the success of the operation.
+   */
   open fun addNotification(tripId: String, tripNotification: TripNotification): Boolean {
     var a: Boolean = true
     viewModelScope.launch {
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/notification/CreateNotification.kt
```diff
@@ -0,0 +1,140 @@
...
+                  // Logic to create notification
+                  val tripNotification =
+                      TripNotification(
+                          notificationId = "", // modified by database
+                          userId = "", // modified by database
+                          title = title,
+                          userName = "", // modified by database
+                          description = description,
+                          timestamp = LocalDateTime.now())
+                  // if successful, call onSuccess(), otherwise onFailure()
```
Exino329, line 113: For the moment, it is not necessary to define onFailure() and onSuccess() case behaviour, for the moment we assume that after adding the notification in the view Model, we will just navigate back to the notification screen. 
We can assume that if the add of the notif fails, we also navigate back to the Notification screen. (We can change the behaviour lated if required.

app/src/main/java/com/github/se/wanderpals/ui/screens/notification/CreateNotification.kt
Exino329, line 1: Good class overall ! Few things could be probably improved  : 
- It is not necessary to define 3 arguments `onSuccess` `onFailure` `onCancel` because `onCancel` `onSuccess`  have the same purpose : navigating back to the notification screen. It is a good idea to think about a failure `onFailure` but for the moment it is simplier to assume that we don't check for failure from the data base when adding a notification and simply also go back to the notification screen.
- It could be nice to leave an errorMessage to notify which  field entry is not valid in addition to the red borders to provide more information to the user.
- The textfield for the description could be a bit larger (vertically) since we have space available for this screen. 
- It is better if the content (i.e. the two textfields and the button) is closer from the cancel button (i.e. put them higher) to not leave a feeling to the user that the screen is "empty".

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CreateSuggestion.kt
Exino329, line 1: This modification is not related to the current issue. It is not a problem however be careful to only make changes that are addressing the related issue/PR. I just leave this comment as a warning and don't worry I also already did this kind of things :')

app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateNotificationViewModel.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.github.se.wanderpals.model.repository.TripsRepository
+import kotlinx.coroutines.launch
+
+/**
+ * ViewModel for creating notifications. This ViewModel is responsible for adding a notification to
+ * a trip.
+ */
+open class CreateNotificationViewModel(tripsRepository: TripsRepository) : ViewModel() {
+  private val _tripsRepository = tripsRepository
+
```
Exino329, line 15: It could be nice to leave a comment to describe what the function does to improve code readability.

Exino329 (changes requested):

app/src/main/java/com/github/se/wanderpals/ui/screens/notification/CreateNotification.kt
annhit29: Okay! And if required, we can still modify it later :D

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CreateSuggestion.kt
annhit29: Thank you for the reminder!

app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateNotificationViewModel.kt
annhit29: Yes, I added it! Please have a check :D

app/src/main/java/com/github/se/wanderpals/ui/screens/notification/CreateNotification.kt
annhit29: Thank you for your precious review!
1) I changed both `onSuccess` and `onCancel` to `onNavigationBack`. I kept `onFailure` and we discuss what we will do later for this.
2) Yes, I added them: "Title cannot be empty" and "Description cannot be empty" for the trip notification title and description respectively. Moreover, the word counter and the outlinedTextField of the title change the color from red back to gray when there is a letter.  The outlinedTextField of the description works in the same way.
3) Yes, I changed the OutlinedTextField of the description to `height(450.dp)`.
4) Yes, this is done by enlarging the OutlinedTextField of the description :D

Exino329 (approved):

Issue #156, opened 2024-04-12
Admin Notification delete


tosigame:
Admins should be able to delete an existing notification.

PR #181, opened 2024-04-19
Admin Notification delete

annhit29:
In this pull request,
I created the NotificationViewModel.kt in which the `deleteTripNotification()` function deletes a trip notification if the current user is an administrator, because. I check if the current user has permission to remove items with the use of `SessionManager.canRemove()` function (only admins can delete notifications).

This is for the issue #156.


tosigame (approved):
Looks good, congrats on the 100%line coverage, well documeted and the tests cover the basic test cases. LGTM
:shipit:

PR #185, opened 2024-04-19
Fixed Data classes to have a time

tosigame:
Added Time parameter for Suggestion and Comment data classes


ForestDope (approved):
Nothing special too say other than good work ! Must have been painful to change every test. The documentation was updated too, nice.

PR #186, opened 2024-04-19
Added a member list and detail screen

AgeX21:
In this pull request, the following has been done : 

- There is now a screen for viewing the list of members in the trip : their name and role
- There is a popup card that displays more information about the user


ForestDope (changes requested):
There should be a separate view model for the members list screen and not use the dashboard one. Some documentation on the functions would be nice too.

ForestDope (approved):
Required changes were correctly implemented, good work !

PR #187, opened 2024-04-20
Refactored NavAction 

mtv47:
Instead of making a single NavHost, I've reworked the NavAction to include both of the needed NavControllers for the two present NavHost. The back button works now because of the code written in this commit because the NavGraphs of both NavHosts are now connected. 

Variables in NavAction are keeper here after asking for the groups advice. The code is more readable this way and since most of the screens need to go throw Route.Trip because of the navigation bar, passing parameters with routes would have been cumbersome

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/MainActivity.kt
+++ b/app/src/main/java/com/github/se/wanderpals/MainActivity.kt
@@ -157,7 +157,9 @@ class MainActivity : ComponentActivity() {
                         Route.DASHBOARD
                       }
                   val tripId = navigationActions.variables.currentTrip
-                  Trip(navigationActions, tripId, tripsRepository, placesClient, routeToGo)
+                  navigationActions.tripNavigation.setNavController(rememberNavController())
+                  navigationActions.tripNavigation.setStartDestination(routeToGo)
+                  Trip(navigationActions, tripId, tripsRepository, placesClient)
                 }
                 composable(Route.CREATE_TRIP) {
                   CreateTrip(OverviewViewModel(tripsRepository), navigationActions)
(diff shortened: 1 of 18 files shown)


mtv47:
Not initialising "@RelaxedMockK lateinit var mockNavActions: NavigationActions" in tests makes the heap fill up. I don't understand why but after speeding too much time I found the where it came from

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -118,83 +118,76 @@ class MainActivity : ComponentActivity() {
...
+                      if (navigationActions.variables.suggestionId != "") {
+                        Route.SUGGESTION
+                      } else if (navigationActions.variables.currentAddress != "") {
+                        Route.MAP
+                      } else {
+                        Route.DASHBOARD
+                      }
+                  val tripId = navigationActions.variables.currentTrip
+                  navigationActions.tripNavigation.setNavController(rememberNavController())
+                  navigationActions.tripNavigation.setStartDestination(routeToGo)
```
Polo150202, line 161: Instead of modifying the navactions start destination, you can directly create a variable in the navactions that remembers which navcontroller was used so that, when you go back, you know which graph was used.

app/src/main/java/com/github/se/wanderpals/ui/navigation/NavigationActions.kt
```diff
@@ -1,120 +1,157 @@
...
+import androidx.navigation.compose.rememberNavController
 import com.github.se.wanderpals.model.data.GeoCords
 
-/**
- * Class defining the navigation actions in the app.
- *
- * @param navController The navigation controller.
- */
-class NavigationActions(private val navController: NavHostController) {
+@Composable
```
Polo150202, line 13: You can maybe add documentation

app/src/main/java/com/github/se/wanderpals/ui/navigation/NavigationActions.kt
```diff
@@ -1,120 +1,157 @@
...
-   * @param route The route to navigate to.
+   * @return The current destination.
    */
-  fun navigateTo(route: String) {
-    navController.navigate(route) {
-      popUpTo(navController.graph.findStartDestination().id) { saveState = true }
-      launchSingleTop = true
-      restoreState = true
+  @Composable
+  fun getCurrentDestination(): NavDestination? {
```
Polo150202, line 38: You can delete the used function

Polo150202 (changes requested):
Great Job with just minor changes

Polo150202 (approved):
Nice job

PR #188, opened 2024-04-21
Fix trips notification data class and create the annoucement data class with some fixes

tosigame:
tranformed the previous Tripnotification into the announcements
and created a new TripsNotification, which correspond better to the wanted behavior.
added/fixed existing tests.


Exino329 (approved):
The large amount of code is due to the refactor of the data class TripNotification and the creation of the new data class Announcement but this PR is mostly doing refactoring. Everything is ok !

PR #189, opened 2024-04-21
Notification screen : UI

Exino329:
The goal of this pull request was to make the UI for notification Screen. At the top, there is a menu with 2 buttons that the user can click to either display the list of notifications or the list of announcement.
- For a notification , its title, and the date / hours at which it has been created is displayed
- for an announcement, the display is the same as for notifications, with in addition the userName of the person who made the announcement. 
- A button for creating an announcement is also displayed if the user selects the announcement (does not depend if the user is an admin or not for the moment)

This pull request only relies on UI design and not all the logic which will be done in sprint#5


app/src/androidTest/java/com/github/se/wanderpals/notifications/NotificationsTest.kt
mtv47, line 1: As discussed, since this PR is only about the UI without any logical functionality, tests will be added later on.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/notifications/Notification.kt
```diff
@@ -0,0 +1,256 @@
...
+                                      else FontWeight.Normal))
+                    }
+              }
+        }
+
+    HorizontalDivider(color = Color.Black, thickness = 2.dp, modifier = Modifier.fillMaxWidth())
+    LazyColumn(modifier = Modifier.fillMaxSize().weight(1f)) {
+      val itemsList = if (notificationSelected) notificationsList else announcementList
+      items(itemsList) { item ->
+        when (item) {
```
mtv47, line 111: Great use of when functionality

mtv47 (approved):
Great looking graphics, nothing to add since the functionality will be added later. Excellent Job!

PR #190, opened 2024-04-21
Added Anon Sign In and Refactored Email SignIn

mtv47:
Used Firebase to create an anonymously sign in option, and refactored the email sign in to have a password and work with Firebase as well


Polo150202 (approved):
Nice code

PR #191, opened 2024-04-21
Add a button to delete a suggestion

AgeX21:
This PR features the following : 

- There is now a button to delete a suggestion in the suggestion feed, only admins or the suggestion's owner can delete a suggestion.


ForestDope (approved):
Looks good to me !  (Good reuse of existing similar code for consistency)
Also nice that you kept the isBottomSheetVisible variable and didn't create a new one as it would have been useless.

PR #211, opened 2024-04-22
Fixing flickering and ViewModel problems

tosigame:
This PR introduces inner factory classes within our ViewModel classes. The primary aim of this architectural adjustment is to enhance the creation and management of ViewModel instances across different screens, especially in cases where multiple ViewModels of the same type are needed but with minimal lifecycle dependency and reduced redundancy in data fetches.
Also fixed Overview, to load the trips when entering the UI (called only once)


mtv47 (approved):
The app is so fluid now, Great catch!

PR #215, opened 2024-04-24
Fixed a small bug inside of BottomNavBar when Back Button is pressed

mtv47:
Moved the MutableStateFlow from the BottomNavBar to the NavigationAction, updated it when the back button is pressed so that the icon below the trip gets to be displayed correctly, the override inside of MainActivity isn't pretty but it was the simplest option that I have found, other solutions didn't trigger every time for example.


Exino329 (approved):
This PR only does small changes so there is not a lot to said on it. Everything looks ok to me.

PR #218, opened 2024-04-25
See details about already present suggestions and stops

mtv47:
In this PR , I worked on displaying information about already present suggestions and stops in the bottom sheet in the same way as a searched destination would. 

This required some rewrites to store the google place id that wasn't keeper before.


Polo150202 (approved):
Great Job

PR #219, opened 2024-04-25
Refactor create suggestion

AgeX21:

- CreateSuggestion has been refactored to use a suggestion as an argument instead of individual fields, it will be necessary to edit suggestions later on
- Other files have been modified to accommodate the change


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
```diff
@@ -284,10 +286,22 @@ fun Map(
                 state = markerState,
                 title = "Click to Create Suggestions",
                 onInfoWindowClick = {
```
tosigame, line 288: Move setVariablesSuggestionId and Location to setVariablesSuggestion since they are always set when setVariablesSuggestion is called

app/src/main/java/com/github/se/wanderpals/ui/navigation/NavigationActions.kt
```diff
@@ -162,10 +163,19 @@ data class NavigationActions(
...
+  fun setVariablesSuggestionId(suggestionId: String) {
     variables.suggestionId = suggestionId
   }
 
+  /**
+   * Set the variables for the navigation actions.
+   *
+   * @param suggestion The suggestion.
+   */
+  fun setVariablesSuggestion(suggestion: Suggestion) {
```
tosigame, line 175: Look at the map comment.

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CreateSuggestion.kt
tosigame, line 1: The change to use the Suggestion as argument, improves code readability 👍 
Also this removes a code smell, which is always 🥇

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
tosigame, line 1: apart from the fix, 👍

tosigame (changes requested):
Overall Great job, the code became more readable, and easier to maintain, however you should change the commented sections, to avoir having conflicts with previous merges. 
👍

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
tosigame: Nice!

tosigame (approved):
Great work, and thanks for being quick! :crossed_swords: :shipit:

PR #221, opened 2024-04-26
Transform Suggestion into Stop

AgeX21:
In this PR, the following was done:

- The agenda now only displays the accepted stop
- Admins can now transform a suggestion into a stop


app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionBottomSheet.kt
annhit29, line 1: This is good work.
- in line39: You can remove the `@param suggestion`'s in the comment as there is no suggestion parameter in the SuggestionBottomSheet's signature.
- In the AlertDialog, the "Confirm Deletion", "Are you sure you want to delete this suggestion?" and "Confirm" can be in red, which represents the user (in our case the administrator of the trip) is doing potential danger actions.

Note that for now, when the administrator of the trip clicks on the "Transform suggestion to a stop" button of the ModalBottomSheet, the suggestion is indeed transformed into a stop that we can see in the agenda! yet we need to go to the dashboard and then go back to the suggestion page to make that transformed suggestion item disappear on the suggestion page. It is an issue we will fix in sprint 6.
When the administrator of the trip clicks on the "Delete suggestion" then click on "Confirm", it is deleted automatically on the suggestion page without the need of going to the dashboard then come back to the suggestion page.

app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
annhit29, line 1: Good!

app/src/main/java/com/github/se/wanderpals/model/viewmodel/AgendaViewModel.kt
annhit29, line 1: This is good too.

annhit29 (changes requested):
The line coverage is 100%, and all the tests have passed on my end!
The code is in overall great. Please look at the comments for SuggestionBottomSheet.kt.

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionBottomSheet.kt
annhit29, line 1: - The comment for `@param suggestion` has been removed, good.
- Make only the "Confirm" button red is also good enough. This was linked to the PR #191 .

annhit29 (approved):
Everything is good now.

PR #222, opened 2024-04-26
Edit Suggestion

AgeX21:

- It is now possible to edit an already existing suggestion if the user has the right to do so.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateSuggestionViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateSuggestionViewModel.kt
@@ -22,6 +22,14 @@ open class CreateSuggestionViewModel(tripsRepository: TripsRepository) : ViewMod
     return a
   }
 
+  open fun updateSuggestion(tripId: String, suggestion: Suggestion): Boolean {
+    var a: Boolean = true
+    viewModelScope.launch {
+      _tripsRepository.updateSuggestionInTrip(tripId, suggestion).also { a = it }
+    }
+    return a
+  }
+
   class CreateSuggestionViewModelFactory(private val tripsRepository: TripsRepository) :
       ViewModelProvider.Factory {
 
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateSuggestionViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateSuggestionViewModel.kt
@@ -23,11 +23,11 @@ open class CreateSuggestionViewModel(tripsRepository: TripsRepository) : ViewMod
   }
 
   open fun updateSuggestion(tripId: String, suggestion: Suggestion): Boolean {
-    var a: Boolean = true
+    var success: Boolean = true
     viewModelScope.launch {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateSuggestionViewModel.kt
```diff
@@ -22,6 +22,14 @@ open class CreateSuggestionViewModel(tripsRepository: TripsRepository) : ViewMod
     return a
   }
 
+  open fun updateSuggestion(tripId: String, suggestion: Suggestion): Boolean {
+    var a: Boolean = true
```
Exino329, line 26: `a` is not explicit. Maybe something as "success" could be better to describe what the variable represent

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CreateSuggestion.kt
```diff
@@ -335,14 +335,27 @@ fun CreateSuggestion(
                                   website = _website,
                                   imageUrl = ""))
                   // Pass the created suggestion to the callback function
-                  if (viewModel.addSuggestion(tripId, suggestion)) {
-                    onSuccess()
+                  if (suggestion.suggestionId != "") {
```
Exino329, line 338: Note : you can use `is(Not)Empty` method to check if a string is empty/not empty.

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CreateSuggestion.kt
```diff
@@ -335,14 +335,27 @@ fun CreateSuggestion(
...
+                      onSuccess()
+                    } else {
+                      onFailure()
+                    }
                   }
                 }
               }) {
-                Text("Create Suggestion", fontSize = 24.sp)
+                Text(
+                    if (suggestion.suggestionId == "") "Create Suggestion" else "Edit Suggestion",
```
Exino329, line 357: As said before, you can also use `IsEmpty` method.

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionBottomSheet.kt
```diff
@@ -40,7 +42,10 @@ import com.github.se.wanderpals.service.SessionManager
  */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun SuggestionBottomSheet(viewModel: SuggestionsViewModel) {
+fun SuggestionBottomSheet(
```
Exino329, line 45: Documentation for the function  `SuggestionBottomSheet` could be added to improve code readability.

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CreateSuggestion.kt
Exino329, line 1: Just a little remark : care with `if` `else` nesting between lines 338-352. But it seems that here we can't do in an another way to improve readability.

Exino329 (changes requested):

Exino329 (approved):

PR #223, opened 2024-04-26
Edit comment

AgeX21:

- It is now possible to edit a comment if the user has the right to do so.


app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionDetail.kt
```diff
@@ -289,6 +283,59 @@ fun SuggestionDetail(
...
+                                // text is empty
+                              }
+                            }
+                          }
+                    },
+                    keyboardOptions = KeyboardOptions.Default.copy(imeAction = ImeAction.Done),
+                    keyboardActions =
+                        KeyboardActions(
+                            onDone = {
+                              if (editingComment) {
```
tosigame, line 315: You duplicated this line, (see Sonar cloud), you should use a function and pass it around instead of duplicating the code! This will make sonarcloud happy!
if (editingComment)...

app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
tosigame, line 1: Looks good, good job, and thanks for adding documentation, it makes it easier to understand :+1:

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/CommentOptionsBottomSheet.kt
tosigame, line 1: :+1:

app/src/main/java/com/github/se/wanderpals/ui/screens/suggestion/SuggestionDetail.kt
tosigame, line 1: Apart from decreasing code duplication, everything else seems good to me! Nice :fire:

tosigame (changes requested):
Overall well done, just try to decrease the code duplication, to keep sonar cloud happy!

tosigame (approved):
Great, This greatly decreased code duplication, also the coverage is lower than expected, because we have the SonarCloud bug which doesn't account some of the files line coverage properly. Good job 🚢

PR #255, opened 2024-04-27
App theme updated 

ForestDope:
Updated the app theme. Making a PR now so that other members can use the theme properly. In the future existing hex value for colors in the code should be replaced with appropriate colors from the color scheme.


AgeX21 (approved):
LGTM ! Nice change for future development

Issue #203, opened 2024-04-22
Transforms the suggestion voting system at the stop


Polo150202:
nbDeLikes > ceil(nbDePersonnesDansLeTrip/2)   ==> add suggesiton to stop et niveau UI suggestion becomes semi-transparent, and therefore suggestion is no longer modifiable

PR #256, opened 2024-04-28
Suggestion to stop transformation based on the number of likes

annhit29:
In this pull request:

We transform a suggestion to a stop based on the following conditions:
    - If the number of users is odd, the threshold of likes is the middle user to ensure the majority (for example: if there are 3 users in a trip, then the threshold of likes is 2, thus the suggestion with 2 likes becomes a stop of the trip).
    - If the number of users is even, the threshold of likes is the middle two users (one extra) to ensure the majority (for example: if there are 4 users in a trip, then the threshold of likes is 2 **+1**, thus the suggestion with 3 likes becomes a stop of the trip).

- If the condition is met without the current user's action, clicking on the heart icon, the suggestion is transformed automatically. Thus in this case, the current user will not see the suggestions which have already met the condition in the suggestion screen, because they are already in the agenda screen in which stops are listed.
This is done by calling `checkAndAddSuggestionAsStop` function on each `suggestion` of the suggestion list in the function `loadSuggestion` of `SuggestionsViewModel`.
- Otherwise, the transformation requires the current user to click on the heart icon. This suggestion will disappear from the suggestion list and other suggestions that have not been converted into stops and were under that suggestion will move up.

\
This pull request is from issue #203.


app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
```diff
@@ -50,6 +51,10 @@ open class SuggestionsViewModel(
   // item:
   private val _likedSuggestions = MutableStateFlow<List<String>>(emptyList())
 
+  // This will hold the IDs of suggestions that have been added to stops
+  private val _addedSuggestionsToStops = MutableStateFlow<List<String>>(emptyList())
```
Polo150202, line 55: You can maybe merge these two val

app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
```diff
@@ -188,6 +203,43 @@ open class SuggestionsViewModel(
...
   }
 
+  /**
+   * Checks if a suggestion of a trip has reached the majority of likes and adds it as a stop if so.
+   *
+   * @param suggestion The suggestion of a trip to check and add as a stop.
+   */
+  open fun checkAndAddSuggestionAsStop(suggestion: Suggestion) {
+    viewModelScope.launch {
+      val allUsers = suggestionRepository?.getAllUsersFromTrip(tripId) ?: emptyList()
```
Polo150202, line 213: You can remove the database fetch from the checkAndAddSuggestionAsStop so that it isn't called several times in the foreach loop of the loadSuggestion function. So you can put it outside of the function to have only one fetch

Polo150202 (changes requested):
Great jobs with only small changes

app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
annhit29: Yes, thank you!

app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
annhit29: Done, thank you!

Polo150202 (approved):
Nice

PR #257, opened 2024-04-28
Enhancements to Balance and Expense Management

tosigame:
This PR introduces new functionalities for managing balances and expenses within our Firestore setup, including necessary subcollections for each trip document.

Key Features:

Balances: We've added a balances subcollection to store and manage a map of user balances (String to Double) with methods to set and retrieve these balances.
Expenses: The expenses subcollection now supports operations to add, update, retrieve(all or one), and remove expenses from a trip.
Tests:

New unit tests ensure these features work as expected, covering the complete lifecycle of balances and expenses within trips.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
@@ -26,7 +26,6 @@ import java.util.UUID
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.tasks.await
 import kotlinx.coroutines.withContext
-import kotlin.math.exp
 
 /**
  * Provides an interface for accessing and manipulating trip data stored in Firestore. Utilizes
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
Exino329, line 1: As discussed, balance will works like TripNotification i.e. we will fetch/store the map on the database at each access.

app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -256,6 +322,201 @@ class TripsRepository(
...
+   * Adds a new expense to a specified trip. This method generates a unique identifier for the new
+   * expense, stores it in the Firestore, and updates the trip document to include this new expense
+   * ID.
+   *
+   * @param tripId The unique identifier of the trip to which the expense is being added.
+   * @param expense The `Expense` object to be added.
+   * @return The unique identifier of the newly added expense if the operation is successful, or an
+   *   empty string if it fails. The method logs the outcome of the operation.
+   */
+  suspend fun addExpenseToTrip(tripId: String, expense: Expense): String =
```
Exino329, line 404: As discussed, Some people needed the object id immediatly after adding the operation, so now the return type is a string representing the ID of the new object added to firebase in case of succes, else empty string to notifiy an error.

Exino329 (approved):
The methods are nicely documented and the logs allow us to keep track of what's going on with log cat. Great job !

PR #258, opened 2024-04-28
Make stop interactable with Map in Agenda

Exino329:
This PR aims to make stop items interactable and upgrade the stops in agenda view.
- in the agenda view when selecting a specific day containing stops, we can now click on the map icon that will make us navigate to the specific stop. If the stop has nos address, the button will not be enabled and displayed in grey
- The fact that stops doesn't necessary have an address is now handled in the UI of the agenda.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
@@ -311,8 +311,8 @@ fun Map(
                 icon = BitmapDescriptorFactory.defaultMarker(BitmapDescriptorFactory.HUE_AZURE),
                 onInfoWindowClick = {
                   // check if stop.stopId contains a "/" and if it does, split it and get the first
-                  if (stop.stopId.last() != '/' && stop.stopId.contains('/')) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
```diff
@@ -311,8 +311,8 @@ fun Map(
                 icon = BitmapDescriptorFactory.defaultMarker(BitmapDescriptorFactory.HUE_AZURE),
                 onInfoWindowClick = {
                   // check if stop.stopId contains a "/" and if it does, split it and get the first
```
ForestDope, line 313: need to update the comment to match the change of the uid of a stop

ForestDope (changes requested):
Update the comment in Map.kt

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
Exino329: Thank you for the catch !

ForestDope (approved):
Nice work

PR #260, opened 2024-04-28
Fixed User delete/kick proper database management

tosigame:
This PR addresses an issue where a user's associated `TripIds` were not being properly cleaned up when the user was deleted or removed from a trip. To ensure the integrity of our data, the `removeTripId` function has been enhanced to accept an optional `userId` parameter. This allows for flexibility in specifying which user's trip list needs to be modified, defaulting to the current user's ID if none is provided. This fixed the bug for kicking a user, where the kicked user, could still access the trip.

(pleasantly this was a lot easier and faster than expected :D )


ForestDope (approved):
Looks good to me.

PR #262, opened 2024-04-29
Save draft markers on map

mtv47:
* Created a SharedPreferencesManager that can be used by everyone to save instances of classes or variables. It's currently used to save temporary PlaceDatas that were places on the map throw the search but weren't transformed into suggestions. 

* Replaced the OnInfoWindowClick  with OnClick for stops and suggestions
* Reduced Google Place API calls by fetching when the search is clicked instead of when the proposition is cleared

* Added Tests and Adapted already existing ones

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
@@ -361,12 +361,14 @@ fun Map(
         Modifier.align(AbsoluteAlignment.BottomRight)
             .padding(horizontal = 16.dp, vertical = 60.dp)) {
           if (listOfMarkers.isNotEmpty()) {
-            Button(onClick = { mapViewModel.clearAllSharedPreferences() }) {
-              Icon(
-                  imageVector = Icons.Default.Clear,
-                  contentDescription = "Clear",
-                  modifier = Modifier.size(24.dp))
-            }
+            Button(
+                onClick = { mapViewModel.clearAllSharedPreferences() },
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
AgeX21, line 1: Could have adjusted Map to use the new theme. Can be done in a separate PR.

mtv47: Do you mean the colors of the elements such as buttons or the background of the map itself ?

AgeX21 (approved):
LGTM! There's not much to say, you could have added the WanderPals Theme to Map at the same time but it's probably better to do one PR responsible for changing everything. Good job!

mtv47:
> LGTM! There's not much to say, you could have added the WanderPals Theme to Map at the same time but it's probably better to do one PR responsible for changing everything. Good job!

Unfortunately, if your are talking about the map background, it's a bit more involved because it's a separate json with colors, someone that is more design oriented should look into it if possible

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
AgeX21: Buttons seem fine for now but having everything in a WanderPalTheme would be the better approach for the future and would prevent the auto-theme from material and compose. The background color of the map should also match the overall theme of the App (white - blue). Not necessary for now.

AgeX21:
I agree. This should make for a future Issue with a separate PR.

PR #263, opened 2024-04-29
Fix navigation bar to properly display current screen selected

mtv47:
* Made NavAction's bottom route more stable by updating it the simplest way I know how, directly inside of the compostables themselves.

* Tried to make the navigateTo more stable as well by checking and doing nothings if the route is not inside predefined lists

* This PR can be opened up later to fix more issues but should be merged the earliest possible since it adds more logging


Polo150202 (approved):
Great Job

PR #264, opened 2024-04-29
Add notification for stop and handling of notifications in case of deletion

Exino329:
This PR aims to add notifications for stop and handling of notification when updates occurs.
- If a stop has an address, the user can click on the item in notification view making him navigate to the map on the corresponding stop, if it has no address, the notification is not clickable (marked by a grey text for the title)
- If a suggestion is deleted or converted into a stop but is still in the notification view, it will be updated as non clickable in the notification view. (a notification saying that a stop has been added if the suggestion was converted in a stop will also be added in the notification view)
- when the user creates a trip, a notification saying that he joined the trip will now be added.


app/src/main/java/com/github/se/wanderpals/model/data/TripNotification.kt
```diff
@@ -7,12 +7,13 @@ import java.time.LocalDateTime
...
  * @param title A brief title of the TripNotification.
- * @param path A string path that could represent the itinerary path or related resource.
+ * @param route The route of the navigation (can be empty).
  * @param timestamp The exact date and time when the TripNotification was created.
+ * @param navActionVariables The navigation action variables stored as a string.
  */
 data class TripNotification(
     val title: String,
-    val path: String,
+    val route: String,
```
mtv47, line 16: I don't think this change was needed, it makes some files are changing for no reason which makes this PR bigger. Maybe in the future, see if it's really relevant to do it. Just an opinion

app/src/main/java/com/github/se/wanderpals/model/viewmodel/OverviewViewModel.kt
mtv47, line 1: Great job putting it inside of the viewModels and not inside of the Composables

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/notifications/Notification.kt
mtv47, line 1: Great refactoring, it's cleaner

app/src/main/java/com/github/se/wanderpals/service/NotificationsManager.kt
```diff
@@ -67,8 +88,32 @@ object NotificationsManager {
...
+    tripsRepository.setNotificationList(tripId, notifList.toList())
+  }
+
+  /**
+   * Adds a notification for a new stop to the notification list of a trip.
+   *
+   * @param tripId The ID of the trip to which to add the notification.
+   * @param stop The added stop for which to create the notification.
+   */
+  suspend fun addStopNotification(tripId: String, stop: Stop) {
```
mtv47, line 104: Consistent with addCreateSuggestionNotification, very readable

app/src/main/java/com/github/se/wanderpals/service/NotificationsManager.kt
```diff
@@ -38,6 +39,25 @@ object NotificationsManager {
     return notifList.toList()
   }
 
+  /**
+   * Removes the path from a notification containing the provided suggestion ID.
+   *
+   * @param tripId The ID of the trip containing the notification.
+   * @param suggestionId The suggestion for which the notification path should be removed.
+   */
+  suspend fun removeSuggestionPath(tripId: String, suggestionId: String) {
```
mtv47, line 48: Elegant solution

mtv47 (approved):
Very clean code, but could have been smaller if path wasn't replaced by route, can create unnecessary conflicts for example. I'm not the judge if it's relevant to have done it, just keep it in mind for future PRs

app/src/main/java/com/github/se/wanderpals/model/data/TripNotification.kt
Exino329: Yes you are right, is it just that path makes less sens than route so I wanted to refactor it since notification will probably be used later for other purpose (e.g. finance screen) so I decided to already make the refactoring

PR #265, opened 2024-04-30
Reworked ViewModels with refreshable elements

mtv47:
The goal of this PR was to make elements refreshable manually or automatically. Automatically was needed since the change in ViewModels before this PR made some screen not update when changing/updating states, you needed to close the app and go back to it. 

What was done:
* Removed all inits and replaced them with LaunchEffects since the ViewModels aren't recreated each time we navigate, LaunchEffect(Unit) works in the same way but is called each time the composable is called.
* Added a wrapper for lazy columns to make them have a pull to refresh functionality
* Added refresh buttons when the lazy columns are not shown
* Adapted tests
* Adapted SuggestionDetails since it was very unstable and buggy
* Fixed small bugs from Map made visible by refreshing functionality
* Deleted AnnouncementViewModel and CreateAnnouncementViewModel


Exino329:
Edit : Matvey found that notification were not updated when a suggestion is removed using the like system this implied to add only 2 lines in suggestion view model which we have done in this PR.

tosigame:
Fixed error, when spamming the like button, caused the app to overreact and crash, (now you can't toggle the same button many times when clicking the button many times)

app/src/main/java/com/github/se/wanderpals/model/viewmodel/AdminViewModel.kt
tosigame, line 1: Its better to have proper by-remember variables, than only updating the info localy, good job! 👍

app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
tosigame, line 1: We use launched effects now, to ensure proper data loading(each time the screen is loaded), looks good)

app/src/main/java/com/github/se/wanderpals/model/viewmodel/MembersViewModel.kt
tosigame, line 1: 👍

app/src/main/java/com/github/se/wanderpals/model/viewmodel/SuggestionsViewModel.kt
tosigame, line 1: Tested and the suggestions work properly, the code could be cleaner if you rewrote the whole class, however since it wasn't part of your task, you've adapted effectively to the existing codebase 🎖️

app/src/main/java/com/github/se/wanderpals/service/SharedPreferencesManager.kt
tosigame, line 1: Good job spotting that you should delete the local markers (they were overlooked), once the suggestion is created from one of them.

app/src/main/java/com/github/se/wanderpals/ui/PullToRefreshLazyColumn.kt
tosigame, line 1: Good job, at not duplicating the same code into 20different functions!, This makes it a lot cleane, its just an elegant solution and its very easy to use, when creating future screens as well 🥇

app/src/main/java/com/github/se/wanderpals/ui/navigation/NavigationActions.kt
tosigame, line 1: good catch! (when going back from the  suggestion creation screen, it didn't zoom to the correct pin on the map)

app/src/main/java/com/github/se/wanderpals/ui/screens/Admin.kt
tosigame, line 1: Instead of checking the Session manager, we now check the current user, to propage changes on screen, nice fix!

tosigame (approved):
After reviewing the updated code and conducting manual tests on my device, I discovered one minor bug, which was related to a different issue. The fix was straightforward, so I implemented it myself. The refresh features are working effectively and are intuitive to use. The reload function, which activates upon scrolling up, is conveniently abstracted by a wrapper around the lazy list, facilitating its integration into future screens. Several other minor/major bugs have also been addressed in this update. Consequently, Userflow 1 is stable and ready for release as of this PR.

Excellent Job, @mtv47  👊

PR #266, opened 2024-05-01
Finance screen : UI and expense list UI

Exino329:
The goal of this pull request was to create the initial UI for the finance screen, allowing us to implement easier the next features we will add into it.

- The finance screen view is accessible from the dashboard menu by clicking on finance option
- The finance screen is composed of 3 main views : Expenses,Categories and Debts
- Categories UI is already done. If the user is a viewer, the button for adding an expense will not be available.

This PR is only focused on UI (also for categories). Linking with the database will be done in other tasks.


app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
AgeX21, line 1: I understand what you did and why you did it but I think it would have been better to write the correct code with the database even if there's no way of testing it for now. Maybe just call addExpenseToTrip without the screen to test if the code can pull correctly afterward.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/ExpensesContent.kt
AgeX21, line 1: Other than the comment i made, the screen looks good. Good job!

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/FinanceTopBar.kt
```diff
@@ -0,0 +1,130 @@
...
+                  imageId = R.drawable.categories,
+                  isSelected = currentSelectedOption == FinanceOption.CATEGORIES,
+                  onClick = { onSelectOption(FinanceOption.CATEGORIES) })
+              NavigationButton(
+                  text = FinanceOption.DEBTS.toString(),
+                  imageId = R.drawable.balance,
+                  isSelected = currentSelectedOption == FinanceOption.DEBTS,
+                  onClick = { onSelectOption(FinanceOption.DEBTS) })
+            }
+      }
```
AgeX21, line 92: Maybe add the first divider here since every finance screen seems to have it

app/src/main/res/drawable/balance.png
AgeX21, line 1: Same for every other png, I think we are using SVG files for icon... Needs to be discussed with the team but it doesn't pose a problem for me atm.

AgeX21 (changes requested):
Test coverage is good, the codes looks fine and the screen are pretty. Please look at the comment and decide what to do with the feedback.

app/src/main/res/drawable/balance.png
Exino329: Ok we will check it for further sprints, I also think it is not a problem for the moment, We just need to start userflows for finance screen.

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
Exino329: Here my goal was to set up the finance view model as a skeleton for the communication between finance and database to make easier and only focusing my self on UI.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/ExpensesContent.kt
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Composable function for displaying an individual expense item.
+ *
+ * @param expense The expense to display.
+ * @param onExpenseItemClick Callback function for when an expense item is clicked.
+ */
+@Composable
+fun ExpenseItem(expense: Expense, onExpenseItemClick: (String) -> Unit) {
+  HorizontalDivider(color = Color.Gray, thickness = 2.dp, modifier = Modifier.fillMaxWidth())
```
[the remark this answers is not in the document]
Exino329: The horizontal divider is at the top so it creates a separation with the top bar of finance. But you are right that it makes more sens to put it in the lazy column instead !

AgeX21 (approved):
LGTM !

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
AgeX21: Okay okay, looks good

PR #268, opened 2024-05-01
fixed editing comment + fixed admin screen

AgeX21:

- After editing and then deleting a comment, the app won't crash anymore when sending a comment.
- The owner now cannot delete themselves from the trip without giving their role first
- The owner cannot demote themselves anymore without giving their role first
- Users can now promote other users up to their role - 1 and can now demote themselves to any role below them
- Changed the display a bit to accommodate changes: grayed out inaccessible roles when editing, removed some buttons for the owner which he cannot do


mtv47 (approved):
Simple and effective, great job!

PR #270, opened 2024-05-02
Finance widget

AgeX21:

- The dashboard now has a second widget to display some basic information about the current finance state of the trip


annhit29 (approved):
This looks very great! Good job ;D

PR #271, opened 2024-05-02
Refactored Suggestion Widget

AgeX21:

- Updated suggestion widget visuals to be more coherent with the rest of the App.


app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
annhit29, line 1: Very good! One thing that you can change is to use the 24-hour clock (as in the SuggestionDetail.kt) instead of 12-hour clock, for startTime and endTime `.format(DateTimeFormatter.ofPattern("dd/MM/yyyy hh:mm"))`

annhit29 (changes requested):
100% of the line coverage, and all the tests passed on my end too.
Very good in general, just one thing to change (see another comment).

annhit29 (approved):
Well done! The 24-hour clock is now used :D

PR #273, opened 2024-05-02
Bug fixes regarding NavGraphs, Permissions, Quitting trip

mtv47:
This PR addresses three bugs that were found by other members during their own coding:
* Sometimes the NavGraph isn't set yet and when pressing to fast on the NavBar when coming from the overview, you can try to navigate before the navigation was set. Fixed by checking if a call to this graph throws an exception
* When quitting a trip from the Admin Panel, the overview wouldn't sometimes update (I guess not garbage collected) and would still display a clickable trip even tho it shouldn't. Now displays a warning and asks the user to refresh.
* Maps sometimes still asked for permission even though it was granted or denied, now added a check

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/endToEndTests/EndToEndTestMilestone1.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/endToEndTests/EndToEndTestMilestone1.kt
@@ -18,6 +18,7 @@ import com.github.se.wanderpals.screens.CreateTripoScreen
 import com.github.se.wanderpals.screens.OverviewScreen
 import com.github.se.wanderpals.screens.TripScreen
 import com.github.se.wanderpals.service.MapManager
+import com.github.se.wanderpals.service.SessionManager
 import com.github.se.wanderpals.ui.navigation.NavigationActions
 import com.github.se.wanderpals.ui.navigation.Route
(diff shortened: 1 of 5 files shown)


annhit29 (approved):
> Sometimes the NavGraph isn't set yet and when pressing to fast on the NavBar when coming from the overview, you can try to navigate before the navigation was set. Fixed by checking if a call to this graph throws an exception.

The bug with the NavGraph has been resolved.

> When quitting a trip from the Admin Panel, the overview wouldn't sometimes update (I guess not garbage collected) and would still display a clickable trip even tho it shouldn't. Now displays a warning and asks the user to refresh.
It works very well now!
This is how we can do a demo:
1. There's a trip where I'm the owner and other members are in it as well;
2. I click on the star of one member to transfer the role of owner;
3. then I quit the trip by clicking on the cross icon;
4. In this stage, if we go back to the trip overview screen, then the trip I have just quitted is still there, but if I click on it
5. the message will be shown to ask me to refresh the trip overview screen;
6. Once we have refreshed the trip overview screen, the trip I have quitted is gone.

> Maps sometimes still asked for permission even though it was granted or denied, now added a check
When the permission is denied, the app no longer asks the permission each time we go to the Map screen.

Now, all the three permission options can be memorized.

PR #274, opened 2024-05-02
Upload images from the application to firebase

Polo150202:
Create the possibility to upload an Image on firebase Cloud storage and update the URI. Then permit to upload this URI to Firebase Database per User. This is done when a User wants to Update his profile photo (and remain permanent)

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/Admin.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/Admin.kt
@@ -102,6 +102,8 @@ fun Admin(adminViewModel: AdminViewModel, storageReference: StorageReference?) {
 
   var selectedMember by remember { mutableStateOf<User?>(null) }
 
+  var isAlreadyClicked by remember { mutableStateOf(false) }
+
   val rainbowColorsBrush = remember {
     Brush.sweepGradient(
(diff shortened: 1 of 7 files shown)


tosigame (changes requested):
After testing the app, before even looking at the code, I found a bug that after clicking multiple times on the edit image, it loads many screens and then crashes with a null pointer exception, it would be great if you could fix that.

gradle/wrapper/gradle-wrapper.properties
tosigame, line 1: just repush the version with 8.4. because it can break android studio depending on the current people setup

app/src/main/java/com/github/se/wanderpals/ui/screens/Admin.kt
tosigame, line 1: Nice, this also covers bugs where one4 could click many times which caused the app to crash, with a null pointer exception.

tosigame (changes requested):
just change back the version of the graddle wrapper to 8.4, everything else looks good and didn't crash locally!

tosigame (approved):
All good, you had an error, because someone modified the screen at the same time and that version was merged, so you needed to fix the test. Else its all good, Good job :medal_military:

PR #275, opened 2024-05-02
Fixed create suggestion to get the correct username for the suggestion

AgeX21:

- When creating a suggestion, the correct username is given instead of a placeholder


tosigame (approved):
Clean fix, of adding the correct username when cresting a suggestion👍

PR #276, opened 2024-05-02
Trip overview screen UI based on Figma

annhit29:
In this pull request,
I updated the trip overview screen UI based on the Figma mockup User Flow 5 (for issue #250) . This includes thus the search bar `OverviewTopBar.kt`. The `OverviewTrip.kt`, the `OverviewBottomBar.kt` and the `OverviewContent.kt` have been adapted to the Figma mockup User Flow 5 too.

For now, we don't add the images yet. We will put them later in the OverviewTrip boxes. 
The height of the OverviewTrip boxes becomes larger in order to put images inside them.

Note that the search bar in the trip overview should not have a user icon because there is no global user at the moment and no menu icon because it is useless. The search bar's original menu icon has been replaced with the loop icon, and when the search text is not empty, the cross icon is displayed so that the user can erase what he/she is searching.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTrip.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTrip.kt
@@ -33,12 +33,14 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTrip.kt
```diff
@@ -65,6 +65,43 @@ fun Context.shareTripCodeIntent(tripId: String) {
   startActivity(shareIntent)
 }
 
+fun formatTripTitle(title: String, maxLineLength: Int = 12): String {
```
ForestDope, line 68: Nice utility function

annhit29: Thank you :D

PR #278, opened 2024-05-02
Fixed a bug where the owner couldn't update the role of other users

AgeX21:

- Fixed a bug where the owner couldn't update the role of other users


app/src/androidTest/java/com/github/se/wanderpals/dashboard/AdminTest.kt
tosigame, line 1: After fixing the bug you also add a test, to test if the issue persists, Great job!

app/src/main/java/com/github/se/wanderpals/ui/screens/Admin.kt
tosigame, line 1: Great catch!

tosigame (approved):
You found a bug, you fixed it, and you tested if its fixed. Excellent 🥇 
(Sonar cloud is broken, so we can't see the coverage)

Issue #247, opened 2024-04-27
Update suggestions screens based on figma mockups


ForestDope:
The suggestion screen UI should be updated based on figma mockups. Also update the suggestion creation screen.

PR #281, opened 2024-05-03
Suggestion screen UI and Suggestion creation screen UI, based on Figma Mockup (for issue #247)

annhit29:
In this pull request, I have improved the user experience whenever he/she is on the screens related to suggestions, specifically, by...
1) I adapted the suggestion screen UI. 
2) I moreover updated the suggestion creation screen, which is the screen after clicking on the "+" button.

These improvements are based on the Figma mockup user flow 8. The layout and the styling (namely font, size, color) are based on the Figma mockup user flow 8.

3) The suggestion creation screen is now scrollable. It wasn't before. The test `checkIfCreateSuggestionIsScrollable` in CreateSuggestionTest.kt is created to check that the screen is indeed scrollable.

\
Note that 
- we (the team) have discussed that the search bar is no longer the menu icon, it has been replaced by the loop icon. 
- when the search text field is not empty, the cross icon appears at the right-hand side of the search bar, so that the user can remove what he/she is writing /searching in one click.
- the search bar in the trip overview should not have a user icon because there is no global user at the moment and no menu icon because it is useless.

The search bar is implemented in SuggestionTopBar.kt .

\
\
This pull request is for issue #247


ForestDope (approved):
LGTM

PR #283, opened 2024-05-03
Fixed a bug in suggestion detail 

ForestDope:
Fixed the bug in suggestion detail that made the user face a blank white screen after liking a suggestion in the suggestion detail screen.
To reproduce: 
- Like the suggestion in the suggestion detail 
- The like makes the suggestion transform into a stop


mtv47 (approved):
Effective

PR #284, opened 2024-05-03
Invite user to trip using his app username

mtv47:
In this PR, I've added the functionality of sending an email to a user by clicking an icon on the trip. This creates a Dialog Window where you need to write the username of the user you want to add. If an email isn't available for this user, the dialog does nothing and the email can't be sent. If the username exists, it enables a button which can be pressed, which then executes an Intent that opens an emails already filled out ready to be sent. 

The username is generated with picking the value before the "@" and adding "@gml" for gmail sign in, or "@mleXXX" for email sign in. With XXX being a modulo number of a hashcode.

The username is also now displayed on the Admin Page

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/overview/OverviewTest.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/overview/OverviewTest.kt
@@ -135,10 +135,14 @@ class OverviewTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSu
       shareTripButton1 { assertIsDisplayed() }
       shareTripButton2 { assertIsDisplayed() }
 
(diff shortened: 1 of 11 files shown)


Exino329 (approved):
Good work ! It could be nice to leave an error message when the username is invalid. However since we will udpate the UI for this dialog later. We can do it at this time and not now. Also when clicking on the buttons to invite, the visual effect shows that they are "both clicked" even if only one is selected, this could be also fixed.

Exino329 (changes requested):
When clicking on the buttons to invite, the visual effect shows that they are "both clicked" even if only one is selected, this could be also fixed. I noticed it after accepting initially, that is my bad for this.

Exino329 (approved):
Everything is clean now ! Nice job

PR #285, opened 2024-05-03
Adjusted the display in finance

AgeX21:

- Adjusted the bottom text in finance to display the intended information: the current expense of the user and the total expense of all users.
- The expense screen now correctly updates when creating an expense


Exino329 (approved):
LGTM !

PR #286, opened 2024-05-03
Push notifications

Polo150202:
Create the class for the notification in the app. Creation of a class that extend the firebase cloud messaging class for receiving the notification sent by firebase and display them on the device by formatting them (Overwritting of the onReceivedMessage). Creation of the token to target the correct user device and renew them when necessary. The big issue with this task was to receive test notifications from firebase, in fact I only recently found out that, even though I had firebase owner accreditations, I couldn't test or send notifications from firebase, let alone receive them, as this was due to the use of the epfl email account which prevented me from being able to delete all these modifications. That's why I wasn't able to complete my task, so I could finish it during sprint 7.


app/src/main/java/com/github/se/wanderpals/model/viewmodel/MainViewModel.kt
```diff
@@ -74,6 +74,10 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
     runBlocking { tripsRepository.addEmailToUsername(getUserName(email), email) }
   }
 
+  fun setUserToken(token: String) {
```
ForestDope, line 77: What is this supposed to do ? Doesn't seem to set a token.

app/src/main/java/com/github/se/wanderpals/model/viewmodel/CreateSuggestionViewModel.kt
```diff
@@ -20,6 +20,12 @@ open class CreateSuggestionViewModel(tripsRepository: TripsRepository) : ViewMod
       runBlocking {
         val newSuggestion = _tripsRepository.getAllSuggestionsFromTrip(tripId).last()
         NotificationsManager.addCreateSuggestionNotification(tripId, newSuggestion.suggestionId)
+
+        /*for (userToken in SessionManager.getListOfTokensTrip()) {
```
ForestDope, line 24: Remove if not useful

app/src/main/java/com/github/se/wanderpals/service/NotificationDevice.kt
ForestDope, line 1: Nicely commented file

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -276,7 +312,12 @@ class MainActivity : ComponentActivity() {
                               OverviewViewModel.OverviewViewModelFactory(
                                   viewModel.getTripsRepository()),
                           key = "Overview")
+                  Log.d("CREATE_TRIP", "Create Trip")
+
                   CreateTrip(overviewViewModel, navigationActions)
+                  // create notification to send to the list of tokens
```
ForestDope, line 318: remove comment

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -314,6 +355,28 @@ class MainActivity : ComponentActivity() {
       }
     }
   }
+
+  private fun showNotification() {
```
ForestDope, line 359: You can remove this private function as it was for tests purposes.

ForestDope (changes requested):
Overall nice work, some small changes (mostly code cleanup) to do and it will be good.

ForestDope (approved):
Good work !

PR #307, opened 2024-05-06
Fixed Trip deletion to remove a trips subcollection when its being deleted

tosigame:
This PR updates the deleteTrip method to asynchronously delete all associated subcollections (stops, suggestions, announcements, expenses, users adn ect...) along with the trip itself. Each deletion task is launched within a coroutineScope to ensure that all related components are deleted before the trip document is removed from Firestore. This approach enhances the robustness of the deletion process and ensures a cleaner database state post-deletion. Additional logging has been added for better traceability and error handling is incorporated to manage potential exceptions effectively.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
@@ -24,6 +24,8 @@ import com.google.firebase.firestore.FirebaseFirestore
 import com.google.firebase.firestore.toObject
 import java.util.UUID
 import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.tasks.await
 import kotlinx.coroutines.withContext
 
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/github/se/wanderpals/repository/TripsRepositoryTest.kt
```diff
@@ -68,7 +136,7 @@ class TripsRepositoryTest {
 
         withTimeout(10000) {
           assertFalse(repository.addTripId(nonExistentTripId))
-          assertFalse(repository.removeTripId(nonExistentTripId))
+          assertTrue(repository.removeTripId(nonExistentTripId))
```
mtv47, line 139: Why did this change to return true now ?

app/src/main/java/com/github/se/wanderpals/model/repository/TripsRepository.kt
```diff
@@ -1353,8 +1355,35 @@ open class TripsRepository(
       withContext(dispatcher) {
         try {
           Log.d("TripsRepository", "deleteTrip: Deleting trip")
-          removeTripId(tripId) // remove the trip from the user
-          tripsCollection.document(tripId).delete().await() // delete a given trip by its tripId
+
```
mtv47, line 1358: Great use of multiple launch calls, didn't know it was possible. Nice that it's concurrent.

mtv47 (approved):
Small question but all good otherwise

app/src/test/java/com/github/se/wanderpals/repository/TripsRepositoryTest.kt
tosigame: The previous version of removeTripid, failed if it couldn't find the TripId,which we wished to delete, however this is not the case in mutliple other methods, which for consistency will return true, if the target of deletion doesn't exist. So I just fixed this method to follow the others. Hope this clears things out)

mtv47: Thx for the very clear explanation

PR #308, opened 2024-05-07
Bypass sign in when starting the app and logout functionality

mtv47:
In this PR:
* Added a check to see if a previous user has already signed in, if yes, directly start at Overview with the variables correctly set.
* UID inside of TripRepository can now be changed, this is done because sometime the user logs out but sign ins right after with a different uid, recreating the class doesn't work
* Added a default profile picture for accounts that don't have any
* Added a profile picture inside of Overview, when clicked, displays a prompt asking you if you want to log out

What is fixed:
* Fixed the name of a user that signs in with a vanilla email not having a name because displayName was empty
* Updating the user throw TripsRepository didn't set all variables, now fixed
* Uploading a profile picture to the FirebaseUser as well when changed
* Fixed sometime wrong path was uploaded inside of the profile picture

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/MainActivity.kt
+++ b/app/src/main/java/com/github/se/wanderpals/MainActivity.kt
@@ -161,10 +161,15 @@ class MainActivity : ComponentActivity() {
 
             SessionManager.setUserSession(
                 userId = currentUser.uid,
-                name = currentUser.displayName ?: "",
+                name =
+                    currentUser.isAnonymous.takeIf { it }?.let { "Anonymous User" }
+                        ?: currentUser.displayName
+                        ?: "",
                 email = currentUser.email ?: "",
                 profilePhoto = currentUser.photoUrl.toString(),
-                nickname = viewModel.getUserName(currentUser.email ?: ""))
+                nickname =
+                    currentUser.isAnonymous.takeIf { it }?.let { "Anonymous User" }
+                        ?: viewModel.getUserName(currentUser.email ?: ""))
           }
 
           NavHost(
--- a/app/src/main/java/com/github/se/wanderpals/MainActivity.kt
+++ b/app/src/main/java/com/github/se/wanderpals/MainActivity.kt
@@ -153,10 +153,23 @@ class MainActivity : ComponentActivity() {
                   mainNavigation = rememberMultiNavigationAppState(startDestination = ROOT_ROUTE),
                   tripNavigation =
                       rememberMultiNavigationAppState(startDestination = Route.DASHBOARD))
+          val currentUser = FirebaseAuth.getInstance().currentUser
+          val startDestination = if (currentUser != null) Route.OVERVIEW else Route.SIGN_IN
+          if (currentUser != null) {
+            Log.d("MainActivity", "User is already signed in")
+            viewModel.initRepository(currentUser.uid)
+
+            SessionManager.setUserSession(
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/se/wanderpals/service/SessionManager.kt
```diff
@@ -35,17 +48,31 @@ object SessionManager {
...
       email: String = currentUser?.email ?: "",
       role: Role = currentUser?.role ?: Role.VIEWER,
       geoCords: GeoCords = currentUser?.geoCords ?: GeoCords(0.0, 0.0),
-      profilePhoto: String = currentUser?.profilePhoto ?: "",
+      profilePhoto: String = currentUser?.profilePhoto ?: default_profile_photo,
       tripName: String = currentUser?.tripName ?: "",
       nickname: String = currentUser?.nickname ?: ""
   ) {
-    currentUser = SessionUser(userId, name, email, role, geoCords, profilePhoto, tripName, nickname)
+    val profilePhotoUse = profilePhoto.isNullOrEmpty().let { default_profile_photo }
```
tosigame, line 62: val profilePhotoUse = if (profilePhoto.isNullOrEmpty()) default_profile_photo else profilePhoto
This should at least partially fix the issue, of the picture not displaying. However even with this, it is still inconsistent, and breaks.

tosigame (changes requested):
Hi, after reviewing your code, it seems fine however when testing manually the app, there are 2 bugs taht I found, one Is that the name for annonymous users is not properly set, and the other is incosistent profile pictures, that doesn't load. (I suggested a fix whic partially solves the problems, but there are still other issues)

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTopBar.kt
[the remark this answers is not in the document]
mtv47: Thx 👍

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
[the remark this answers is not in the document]
mtv47: Thx for pointing this out, will be fixed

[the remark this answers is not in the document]
mtv47: I will move it inside of the OverViewModel to make it more stable

app/src/main/java/com/github/se/wanderpals/service/SessionManager.kt
mtv47: This unfortunately didn't work, because profilePhoto from firebase isn't actually null but "null"

app/src/main/java/com/github/se/wanderpals/MainActivity.kt
tosigame, line 1: Now the same profile picture is workingacross multiple trips. And is consistent Nice! :medal_military:

app/src/main/java/com/github/se/wanderpals/ui/screens/overview/OverviewTopBar.kt
tosigame, line 1: Works like a charm ! :+1:

tosigame (approved):
Looks good, the logogut functionnality is tested, you quickly addressed my concerns with the image consistency and it works well on my end. The code is clean and readable. 
Good job! :1st_place_medal:

PR #313, opened 2024-05-09
Finance debt

AgeX21:

- The screen finance debt now displays the list of users with their corresponding debt to other users
- The screen also displays the current user debt with other users with the idea of settling debt when clicking on the composable (needs to be done later)


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/DebtContent.kt
Exino329, line 1: Just a remark concerning this class : 
- instead of use `filter` method to find a user by its user id, you can use the `find` method which is more adapted.

Exino329 (changes requested):
Good code and the screen starts to looks good ! However, there is things that should be reconsidered : 

- It would be better to place widgets indicating the debts concerning the current user at the bottom of the debt summary of all users, since When there is not a lot of user, a lot of free space is not use in the debts content so instead of clicking on a user in the debt history to watch his balance, it would be better to put widgets below the debt history to indicate him the balances only concerning him (who he needs to give money back, and who needs to pay him). So we can keep the history but make it non-clickable and put below the informations that concerns only the current user balance.
- It would be also good to have a feature allowing the user to repaid his debt, so that when the user has repaid his debt, he can confirm it in the application which will update the balance accordingly.
- The content of debt should be placed in a lazy column, Since we don't know how many users will be there on the trip, we need to have the possibility to scroll down in the case there is no sufficient place in the screen to display all information. You can take a look at the other screen content of finance option which use LazyColumn to take inspiration.

Exino329 (approved):
Looks good to me !

Exino329 (approved):

PR #314, opened 2024-05-09
Delete expenses

AgeX21:

- Fixed a bug in TripsRepository where the UID of the user creating the expense was substituted with the "owner" of the expense which resulted in wrong calculation and inaccuracy
- Fixed a bug where the expense screen wouldn't update after creating an expense
- Added a dialog to delete an expense when clicking on it


Polo150202 (approved):
Looks good overall

PR #315, opened 2024-05-09
Delete Trip

AgeX21:

- Added a button in the dashboard's menu to delete the trip if the current user is the owner of the trip

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/dashboard/DashboardTest.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/dashboard/DashboardTest.kt
@@ -10,10 +10,12 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.wanderpals.model.data.Category
 import com.github.se.wanderpals.model.data.Expense
 import com.github.se.wanderpals.model.data.GeoCords
+import com.github.se.wanderpals.model.data.Role
 import com.github.se.wanderpals.model.data.Stop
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/github/se/wanderpals/dashboard/DashboardTest.kt
tosigame, line 1: The test suite effectively handles various scenarios around the trip deletion functionality, ensuring that both the display of the deletion dialog and the role-based access controls function as expected. This approach guarantees robustness in user interactions. 👍

app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
tosigame, line 1: Very neat and easy to understand!

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Dashboard.kt
tosigame, line 1: Tested locally and it looks good, and works even better. I tried to test some common UI bugs on your composables and I didn't find any problems in it. Again very clean code, bravo! 🎖️

tosigame (approved):
Very Clean and consice PR, code is very easy to understand, locally everything works just fine, even when I tried to break it. Tests cover the basic functionnality and even some edge cases. The low coverage is linked to the broken sonarcloud which doesn't track some files for some still unknown reasons.

Overall Great Job, and continue like that in the feature!  👍

Issue #300, opened 2024-05-03
Add marker in Agenda for clarity


AgeX21:
The agenda should have a marker on the dates with active stops.

PR #317, opened 2024-05-10
Date marker of the Calendar in Agenda (for issue #300)

annhit29:
In this pull request,

I added the functionality: when there is a stop (because it has been transformed from a suggestion), then there is a marker under the calendar's date. Otherwise, there are no markers under the date of the calendar (because there are no stops on that date).
For this functionality, I created an enumeration class StopStatus which can be either `ADDED` or `NONE` if a stop was added to the trip or not, on a specific date. The reason for the use of an enumeration class is that we might want to define other statuses for stops in the future.
The StopStatus is stored (see FirestoreStop.kt), I also updated the Stop data class for having the StopStatus for every stop, its default value is set at `NONE` as the suggestion has not been transformed into a stop yet.

I also created tests for testing this functionality (see AgendaTest.kt and AgendaViewModelTest.kt files).

This pull request is for the issue #300 .


app/src/main/java/com/github/se/wanderpals/model/data/Suggestion.kt
```diff
@@ -35,5 +36,6 @@ data class Suggestion(
     val createdAtTime: LocalTime = LocalTime.of(0, 0), // Time of creation
     val stop: Stop = Stop(), // Embed the Stop object directly
     val comments: List<Comment> = emptyList(),
-    val userLikes: List<String> = emptyList()
+    val userLikes: List<String> = emptyList(),
+    val stopStatus: CalendarUiState.StopStatus = CalendarUiState.StopStatus.NONE // Default value
```
AgeX21, line 40: Not required, this variable is only accessed in FirebaseSuggestion to store and retrieve it, never accessed again.
It also doesn't make sense to have the suggestion hold info about the stop after the suggestion has been accepted.
Needs to be removed.

app/src/main/java/com/github/se/wanderpals/model/firestoreData/FirestoreSuggestion.kt
```diff
@@ -30,7 +32,9 @@ data class FirestoreSuggestion(
     val stop: FirestoreStop = FirestoreStop(), // Using Firestore-compatible Stop object
     val comments: List<FirestoreComment> =
         emptyList(), // Using Firestore-compatible Comment objects
-    val userLikes: List<String> = emptyList()
+    val userLikes: List<String> = emptyList(),
+    val stopStatus: String =
```
AgeX21, line 36: like said previously, not required. Should use the text of suggestion to mark accepted suggestion like discussed.

app/src/main/java/com/github/se/wanderpals/model/firestoreData/FirestoreSuggestion.kt
```diff
@@ -56,7 +60,9 @@ data class FirestoreSuggestion(
               suggestion.comments.map {
                 FirestoreComment.fromComment(it)
               }, // Convert each Comment to FirestoreComment
-          userLikes = suggestion.userLikes)
+          userLikes = suggestion.userLikes,
+          stopStatus = suggestion.stopStatus.name // Convert enum to string for Firestore
```
AgeX21, line 64: same thing

app/src/main/java/com/github/se/wanderpals/model/firestoreData/FirestoreSuggestion.kt
```diff
@@ -80,6 +86,8 @@ data class FirestoreSuggestion(
             LocalTime.parse(createdAtTime, timeFormatter), // Parse time string back into LocalTime
         stop = stop.toStop(), // Convert FirestoreStop back to Stop
         comments = comments.map { it.toComment() }, // Convert each FirestoreComment back to Comment
-        userLikes = userLikes)
+        userLikes = userLikes,
+        stopStatus = CalendarUiState.StopStatus.valueOf(stopStatus) // Convert string back to enum
```
AgeX21, line 90: same thing

app/src/main/java/com/github/se/wanderpals/model/viewmodel/AgendaViewModel.kt
```diff
@@ -43,14 +43,48 @@ open class AgendaViewModel(
 
   open var selectedDate: LocalDate? = LocalDate.now()
 
+  open val _stopsInfo = MutableStateFlow<Map<LocalDate, CalendarUiState.StopStatus>>(emptyMap())
```
AgeX21, line 46: Looks good. It may be simpler to use a simple Boolean instead of a custom enum but if you have ideas of other states the stop could be in, it can be a good solution.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/agenda/Agenda.kt
```diff
@@ -308,6 +309,13 @@ fun ContentItem(
   // Assuming dayOfMonth is an empty string for empty dates, or add a specific check if possible.
   val isEmptyDate = date.dayOfMonth.isEmpty()
 
+  // Set the marker color based on the stop status
+  val markerColor =
+      when (date.stopStatus) {
+        CalendarUiState.StopStatus.ADDED -> Color.Blue // Stop added
```
AgeX21, line 315: Looks good.

AgeX21 (changes requested):
Overall good solution and implementation, need some changes but good job :

- remove the stopStatus in the Suggestion data model and Firestore

annhit29:
Hi William,

> It may be simpler to use a simple Boolean instead of a custom enum but if you have ideas of other states the stop could be in, it can be a good solution.

The reason why I used enum is because we might want to define other statuses, such as
- the most recent added: RECENT being associated with the green color
- added before: ADDED being associated with the blue color
- not added yet: NONE being associated with the transparent color (i.e. no colors)

It is easier to adjust the code from enumeration classes to boolean, but not vice versa, and enumeration classes give us more possibilities for potential features in the future.

\
> like said previously, not required. Should use the text of suggestion to mark accepted suggestion like discussed.

If you look at the block comment of the Suggestion data class, you can see:
```
 * @param text The main content of the suggestion, detailing what the user is proposing and the
 *   reasons for it.
 ```
That is not relevant to the goal of the created enumeration class.

\
> Not required, this variable is only accessed in FirebaseSuggestion to store and retrieve it, never accessed again.
It also doesn't make sense to have the suggestion hold info about the stop after the suggestion has been accepted.
Needs to be removed.

Actually, I use this enumeration class to keep the historical suggestions. It is used in `checkAndAddSuggestionAsStop` of the SuggestionsViewModel class. The historical suggestions are those accepted but they are not deleted, so it is a solution to have an enum to mark the status.

Note that by changing the Suggestion data class and the FirestoreSuggestion.kt, there are only 4 lines more. These lines barely changes the code. Moreover, as there is the default value `val stopStatus: CalendarUiState.StopStatus = CalendarUiState.StopStatus.NONE // Default value`, we don't need to change the value of `stopStatus` when it is not necessary. So by adding the enumeration class, it will not complexify the code.

AgeX21:
I can see the utility of the enum class and you explained it well. However, It would make more sense for the enum in the suggestion data model to be moved to the stop data model for 1. coherence and 2. simplicity -> Currently, getting the map of Stop to stopStatus would require for each stop to look at each suggestion for the correct stopId and get the stopStatus which is just not a correct way to do.

annhit29:
After a long discussion of whether simply use the `text` field of the Suggestion data class or use the Boolean (such as `isAdded`) instead of an enumeration class for the `stopStatus` of the Suggesion data class, we have come to the conclusion to move the enumeration class to the Stop data class. Thus, I have changed back the Suggestion data class and the FirestoreSuggestion.kt as they were before. And I have done the same thing (following the same logic as how I did for the Suggestion data class and the FirestoreSuggestion.kt ) to modify both the Stop data class and the FirestoreStop.kt.

AgeX21 (approved):
LGTM!

PR #327, opened 2024-05-11
Expenses info display on click

Exino329:
In this PR, the following have been done :
User can now click on an expense in the expense option view in finance screen, displaying detailed information about an expense, including : 
 - title,cost,date and participants related to this expense.
 -  If user has permission higher than viewer, the user can delete the expense. directly in the expense info view.
 
 N.B : initally, deleting action was done  when clicking on an expense item (PR #314) since expense info feature was not available at this print. Now delete action is available in the expense info view.


tosigame (approved):
Very clean code, great coverage. The UI design is very nice, and neat. I didn't manage to break it. Overall good job, and continue that way! :1st_place_medal:

PR #328, opened 2024-05-12
Fix in Trip Creation

tosigame:
Refactored the CreateTrip function to create only a single trip instance, improving efficiency. Transitioned the implementation to use Coroutine scope in the ViewModel, replacing runBlocking to prevent main thread blockage. Updated the OverviewModel to have a stateflow variable tracking the end of the TripCreate operation. Adjusted and fixed unit tests to align with these changes.

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/trip/CreateTripTest.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/trip/CreateTripTest.kt
@@ -425,80 +425,9 @@ class CreateTripTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withCompose
     }
   }
 
-  @Test
-  fun saveTripWorks() = run {
-    ComposeScreen.onComposeScreen<CreateTripoScreen>(composeTestRule) {
-      val vm = CreateTripViewModelTest(TripsRepository("testUser123", Dispatchers.IO))
-      composeTestRule.setContent { CreateTrip(vm, mockNavActions) }
-      step("Open trip screen") {
-        inputTitle {
-          assertIsDisplayed()
-          performClick()
-
-          assertTextContains("Title")
-          assertTextContains("Name the trip")
-
-          performTextClearance()
-          performTextInput("My trip")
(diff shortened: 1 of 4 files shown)


app/src/androidTest/java/com/github/se/wanderpals/trip/CreateTripTest.kt
mtv47, line 1: I don't have any problems here in the removal of @Before since there isn't many tests present in this file, but I would suggest taking a look at https://stackoverflow.com/questions/13017097/exclude-individual-test-from-before-method-in-junit/13017452#13017452 to see how it could have been done cleaner. You have removed confirmVerified by accident I suppose, but from my research it isn't needed here since there is only one call verify.

app/src/main/java/com/github/se/wanderpals/model/viewmodel/OverviewViewModel.kt
mtv47, line 1: isAddingTrip isn't used but could be useful for other methods and since _isAddingTrip needs to exist anyway, it isn't a big difference

app/src/main/java/com/github/se/wanderpals/ui/screens/CreateTrip.kt
mtv47, line 1: Follows previously written code when a runBlocking is replaced with a viewModelScope.launch, nothing to add

mtv47 (approved):
This PR is a great example for others on how to use spyk for future testing. Great addition

app/src/main/java/com/github/se/wanderpals/model/viewmodel/OverviewViewModel.kt
tosigame: Good point

app/src/androidTest/java/com/github/se/wanderpals/trip/CreateTripTest.kt
tosigame: Great to know! thanks

PR #330, opened 2024-05-14
Live location in map and sharing

mtv47:
In this PR, I've added the functionality of live locations on the map. When pressing a button, it will check and update your position on the app every couple of seconds when the tracking functionality is enabled. This is done as a foreground task on the device, with an additional notification informing the user that his current location is being tracked. 

Fixed a camera bug, when the map was updated, the camera returns to the default 10f zoom, which it doesn't now.

Replaced the icons for the location of users for higher definition ones

I've tried for a very long time to test the Intent or the foreground task, but I wasn't able to do anything. I have absolutely no clue how to test this and haven't found anything on the internet


app/src/main/java/com/github/se/wanderpals/service/LocationServices.kt
tosigame, line 1: Works Great, however you should disable location services, once you exit a trip.

app/src/main/java/com/github/se/wanderpals/service/MapManager.kt
tosigame, line 1: LGTM 👍

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
tosigame, line 1: Very clean new map marker, its more clear now.

tosigame (changes requested):
Overall Great, the code is a bit hard to understand because it introduces new non basic functionnalities, however locally it works great. However you should disable the location services.

app/src/main/java/com/github/se/wanderpals/service/LocationServices.kt
mtv47: Thx for pointing this out, I will do as you say but also disable it when the app is destroyed

tosigame (approved):
Great works, good use of the onDestroy android lifecycle method.Good job this was a hard task, that required to implement new functionnality. 🎖️

PR #331, opened 2024-05-14
Add expense notification and update stop notification

Exino329:
In this review, the following things have been done : 
- When an expense is created an expense notification is created. When clicking, it displays the information of the expense and if the expense is deleted, we can't click on it anymore
- Since there is a stop list implemented now, every stop notification is clickable and navigates to the stop list. The notification text indicates the starting date of the stop so that user can find easier which stops it corresponds to in the notification List
- Tests have been modified and updated to reflect all of these changes


mtv47 (approved):
Great PR, not easy working on notifications since a lot of files interact. A small improvement for consistency can be done, but it's really not mandatory. Great job!

PR #337, opened 2024-05-17
Refactor Suggestion Widget

AgeX21:

- Refactored the suggestion widget in the dashboard to be more in line with the other two widget
- Changed the topbar of the dashboard to be more coherent with the figma and the global theme of the app

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
@@ -3,18 +3,22 @@ package com.github.se.wanderpals.ui.screens.dashboard
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Menu
 import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.ElevatedCard
+import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
```diff
@@ -63,131 +60,169 @@ fun DashboardSuggestionWidget(viewModel: DashboardViewModel, onClick: () -> Unit
...
+                                            color = MaterialTheme.colorScheme.onPrimaryContainer,
+                                            fontWeight = FontWeight.Bold))
+                              }
+
+                          Spacer(modifier = Modifier.padding(4.dp))
+
+                          Text(
+                              text =
+                                  "Total: ${sortedSuggestion.size} suggestion" +
+                                      if (sortedSuggestion.size > 1) "s" else "",
```
Polo150202, line 109: You can maybe use the ternary operator

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
```diff
@@ -63,131 +60,169 @@ fun DashboardSuggestionWidget(viewModel: DashboardViewModel, onClick: () -> Unit
...
+                                  )
+                                }
+                          } else {
+                            Column {
+                              SuggestionWidgetItem(suggestion = sortedSuggestion[0])
+                              HorizontalDivider(
+                                  color = MaterialTheme.colorScheme.surfaceVariant,
+                                  thickness = 1.dp,
+                                  modifier = Modifier.padding(horizontal = 8.dp))
+                              if (sortedSuggestion.size > 1) {
```
Polo150202, line 147: You can use when conidition instead of nested if condition

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Dashboard.kt
```diff
@@ -274,33 +276,41 @@ fun Menu(
...
                 Alignment
                     .CenterVertically // This aligns all children vertically centered in the Row
             ) {
-              ElevatedButton(
-                  modifier = Modifier.testTag("menuButton"),
-                  content = { Icon(Icons.Default.Menu, contentDescription = "Menu") },
-                  onClick = {
-                    scope.launch { drawerState.apply { if (isClosed) open() else close() } }
-                  })
+              Icon(
```
Polo150202, line 288: You can an IconButton instead

Polo150202 (approved):
Nice code overall, but with a lot of "if()" conditions that might increase the complexity

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
AgeX21: It is the ternary operator in Kotlin

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
AgeX21: i can remove a nested if and it would make more sense but switching to a when would not do the same thing and would create a lot of duplication

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Dashboard.kt
AgeX21: I removed the IconButton because it was a bit ugly

Polo150202 (approved):
Nice code

Issue #325, opened 2024-05-11
Voting system with deadline


annhit29:
voting system (new icon with countdown): launch a vote with a deadline of 24 hours for the suggestion. Note that if the number of likes of a suggestion is >50% then it automatically becomes a stop (before the deadline).

PR #338, opened 2024-05-17
Voting system with deadline (for issue #325)

annhit29:
In this pull request, I have done the followings:

1) The admin (Role.ADMIN) or the owner (Role.OWNER) can launch the vote by clicking on the vote icon. Once the admin or the owner has clicked on the vote icon of a suggestion item, this icon can no longer be clicked and the countdown starts. The countdown is displayed in the 24-hour format HH:mm:ss.
If the role is not admin nor owner, then they cannot see the vote icon, but only the remaining time.
2) The 24-hour countdown time is displayed and updated every second in the specified format (HH:mm:ss), which is the remaining time for voting (the "Up" icon) a suggestion.
3) The heart ("Favorite") icon has been replaced by the "Up" icon, because it is closer to the icon for supporting a suggestion.
4) I added two fields in the suggestion data class, and so adapted the FirestoreSuggestion accordingly.
5) Since the icon must be consistent everywhere, so I also replaced the heart ("Favorite") icon with the "Up" icon in the SuggestionHistoryItem.kt .

I thus created some tests for testing the UI part of the SuggestionItem.kt (see SuggestionItemTest.kt) and tests that test suggestionItem-related functionalities in SuggestionsViewModelTest.kt.

\
Note that an "Up" icon is a vote by a User of a trip.

\
This pull request is for the issue #325 .


app/src/main/java/com/github/se/wanderpals/model/firestoreData/FirestoreSuggestion.kt
```diff
@@ -80,6 +104,10 @@ data class FirestoreSuggestion(
             LocalTime.parse(createdAtTime, timeFormatter), // Parse time string back into LocalTime
         stop = stop.toStop(), // Convert FirestoreStop back to Stop
         comments = comments.map { it.toComment() }, // Convert each FirestoreComment back to Comment
-        userLikes = userLikes)
+        userLikes = userLikes,
+        voteIconClicked = voteIconClicked,
+        //            voteStartTime = LocalTime.parse(voteStartTime, timeFormatter) // Parse time
```
Polo150202, line 109: just remove the comment

app/src/main/java/com/github/se/wanderpals/model/data/Suggestion.kt
```diff
@@ -36,4 +40,9 @@ data class Suggestion(
     val stop: Stop = Stop(), // Embed the Stop object directly
     val comments: List<Comment> = emptyList(),
     val userLikes: List<String> = emptyList(),
+    val voteIconClicked: Boolean = false, // By default, the vote icon has not been clicked.
+    // voteIconClicked implies countdownStarted has not started; voteIconClicked is false implies
+    // countdownStarted has started
```
Polo150202, line 45: also here delete the comments

Polo150202 (dismissed):
Nice code overall

app/src/main/java/com/github/se/wanderpals/model/firestoreData/FirestoreSuggestion.kt
annhit29: I just removed it, thank you!

app/src/main/java/com/github/se/wanderpals/model/data/Suggestion.kt
annhit29: Also done here, thank you!

Polo150202 (approved):
Great

PR #340, opened 2024-05-17
Group Document Screen

Polo150202:


annhit29 (changes requested):
Good job for the code!

- You only need to change something for this sprint, such as the use of `MaterialTheme.colorScheme`. 
- And please also comment on your code. 
- Please refer to my comments in the files.
- You should create tests for both `loadLastAddedSharedDocument()` and `loadLastAddedPrivateDocument()` of the DashboardViewModel.kt, and tests for DocumentPSViewModel as well. These tests test the correctness of every function you have implemented in this pull request, and they also increase the line coverage. Please create the tests here:
image
- It would also be great if you could add the description to describe what you have done in this pull request.

\
You can refactor the code in the next sprint.

Everything works well :D

Polo150202:
No problem, I will fix all for the next sprint

annhit29:
Since there are still other issues to do (more urgent for you) for having the functionalities. Adding comments, and changing colors can also be done when you will be refactoring the code and the screen layout and styling. And you will also add tests for the ViewModels in the next sprint.

I will check if everything will have been done in the next sprint :D

annhit29 (approved):
Since there are still other issues to do (more urgent for you) for having the functionalities that work. Adding comments, and changing colors can also be done when you will be refactoring the code and the screen layout and styling, especially because William needs the code for doing the issue #332. And you will also add tests for the viewmodels in the next sprint. Thus, as discussed previously this morning. Paul will finalize all these in sprint9.

I will check if everything will have been done in the next sprint :D

PR #342, opened 2024-05-18
Admin screen overhaul

ForestDope:
In the PR : 
- Admin/Members UI reworked 
- Can now edit name used throughout the trip. 
- Added a visual indicator that edit mode is on (red marks on editable fields). 
- Promoting someone to owner now requires confirmation with a dialog. 
- Fixed a bug that allowed members to demote other members to viewer. 
- Adapted the test suite for the new changes.


app/src/main/java/com/github/se/wanderpals/ui/screens/Admin.kt
AgeX21, line 1: Looks good overall, few things I noticed on my end : 
- The modify option is weird to use, the red color made me think I forgot to do something but it was to signal me I could modify, misleading.
- If the name is too long(2 lines), the email is cut and I don't see the whole email either.
- I don't think there should be a modify Icon next to each user if they don't have the rights to, it's confusing.

AgeX21 (approved):
Looks good to me, please look at my suggestions in admin for later.

PR #343, opened 2024-05-19
Add image to trip

Polo150202:
Create the possibility to add an Image to the trip overview for each trip


AgeX21 (approved):
LGTM ! It works fine but some changes on the UI side would be nice : 
- Add a padding on both side of the card
- Align the buttons and put them on the bottom, maybe use a different color than black to make them pop out more on the Image
- Make a top bar for the card to improve the readability
- Change the default color of the card to be more in theme with the app

Improving the test coverage would be nice too

AgeX21 (approved):
LGTM !

PR #368, opened 2024-05-21
Modify geocords to make less api requests

mtv47:
The goal of this PR was to make less api calls to Google Maps to reduce the apps cost.

* PlaceData was removed and integrated to GeoCords
* The API request isn't made if inside of Firestore, the data is already present
* If the data isn't present inside of Firestore, it's fetched like last time and uploaded to Firestore to not make the request again


app/src/main/java/com/github/se/wanderpals/model/data/GeoCords.kt
tosigame, line 1: All good, here, you correctly gave default values to the parameters, (Else Firebase would panic when trying to push the data). :+1:

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/map/Map.kt
tosigame, line 1: Good way to reduce the number of requests to the Maps Api :fire:

tosigame (approved):
Currently approved, however Github Actions are down (Github side) so we will see if the CI, detects anything which shouldn't be the case, since it locally works for me!
Overall Great job, and this should help us decrease our Maps API usage by a lot. :1st_place_medal:

app/src/test/java/com/github/se/wanderpals/data/GeoCordsTest.kt
tosigame, line 1: Great usage of mockk!

tosigame (approved):
Great, Sonar is still broken so we can't know accurately the real coverage, but the tests cover most of the added functionnality which is great!
Excellent job :firecracker: !

PR #369, opened 2024-05-21
Fix AddAnnouncement to Prevent Duplicate Execution on Multiple Clicks

tosigame:
PR Description:
Ensured AddAnnouncement method does not execute multiple times on multiple clicks.
Changed runBlocking to CoroutineScope.launch to avoid blocking the main thread.
Fixed navigation issues and added LaunchedEffect in CreateAnnouncement screen.
Updated existing tests to reflect these changes.
Added a regression test to simulate multiple clicks and verify only one execution.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/NotificationsViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/NotificationsViewModel.kt
@@ -131,8 +131,6 @@ open class NotificationsViewModel(val tripsRepository: TripsRepository, val trip
         val listOfTokens = tripsRepository.getTrip(tripId)
         if (listOfTokens != null) {
           Log.d("NotificationAnnouncement", "List of tokens: ${listOfTokens.tokenIds}")
-        }
-        if (listOfTokens != null) {
           for (token in listOfTokens.tokenIds) {
             sendMessageToListOfUsers(token, announcement.title)
           }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/wanderpals/model/viewmodel/NotificationsViewModel.kt
```diff
@@ -114,20 +123,32 @@ open class NotificationsViewModel(val tripsRepository: TripsRepository, val trip
...
-      if (listOfTokens != null) {
-        for (token in listOfTokens.tokenIds) {
-          sendMessageToListOfUsers(token, announcement.title)
+    viewModelScope.launch {
+      if (!isAddingAnnouncement.value && !createAnnouncementFinished.value) {
+
+        _isAddingAnnouncement.value = true
+        tripsRepository.addAnnouncementToTrip(tripId, announcement)
+        val listOfTokens = tripsRepository.getTrip(tripId)
+        if (listOfTokens != null) {
```
Polo150202, line 132: You can safely delete this if condition, let only the second one. It will decrease the complexity of the code

Polo150202 (dismissed):
Nice Code

app/src/main/java/com/github/se/wanderpals/model/viewmodel/NotificationsViewModel.kt
tosigame: Done.)

PR #341, opened 2024-05-18
Visual cues for offline mode

mtv47:
In this PR, the following was done when Offline
* "No internet connection" toast on Pull to Refresh
* Added Offline message on screens that had an icon button refresh when Empty List
* Disable Trip Creation and Join Button
* Hide Edit comment
* Hide TransformToStop Suggestion
* Disable Suggestion Creation/Deletion
* Disable/Hide Trip Deletion
* Disable Expense Deletion
* Disable/Hide Expense Creation
* Disable Announcement Creation/Deletion
* Disable Stop Deletion 

Bugs Fixed:
* Fixed SessionManager being online by default when starting the app already Offline
* Announcement Creation Button was hidden when Announcement List contained at least one element

Missing:
* Admin Screen wasn't touched since it is currently being reworked, will be done next week

PR #370, opened 2024-05-22
Offline Admin

mtv47:
As said in "Visual cues for offline mode #341", this PR aims to complete the offline visual cues that were missing in the admin panel last week. 

This is done like the previous PR, by disabling and enabling click actions based on if the device is only or not. 
Additional tests were also added

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/dashboard/AdminTest.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/dashboard/AdminTest.kt
@@ -3,6 +3,7 @@ package com.github.se.wanderpals.dashboard
 import androidx.compose.ui.test.assertCountEquals
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsEnabled
+import androidx.compose.ui.test.assertIsNotEnabled
 import androidx.compose.ui.test.assertTextContains
 import androidx.compose.ui.test.isDisplayed
 import androidx.compose.ui.test.junit4.createComposeRule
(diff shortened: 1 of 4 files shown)


app/src/androidTest/java/com/github/se/wanderpals/dashboard/AdminTest.kt
Exino329, line 1: Just a little note, you can modularize the part of your test that call setContent by creating a function that Set the content for a given user (passed in argument). I just let it here since it can be useful in case we need more tests in the future

app/src/main/java/com/github/se/wanderpals/model/viewmodel/AdminViewModel.kt
```diff
@@ -70,6 +70,7 @@ open class AdminViewModel(
 
   // Modify the current user's profil photo
   open fun modifyCurrentUserProfilePhoto(profilePhoto: String) {
+    if (!SessionManager.getIsNetworkAvailable()) return // Check if the network is available
```
Exino329, line 73: I personally prefer to always but braces in conditional statements but this is just personal preferences.

app/src/test/java/com/github/se/wanderpals/viewmodel/AdminViewModelTest.kt
Exino329, line 1: Looks good to me but what the //Given, //When, // Then comments refers to here?

Exino329 (approved):
Looks good to me ! I just leaved 1-2 comments concerning details but nothing critical needs to be reconsidered.

app/src/androidTest/java/com/github/se/wanderpals/dashboard/AdminTest.kt
mtv47: I find that modularizing the tests themselves can reduce the readability of the tests, here you can clearly see what the tests are doing in my personal opinion. They are small enough in my opinion to not require this

app/src/test/java/com/github/se/wanderpals/viewmodel/AdminViewModelTest.kt
mtv47: It was copied over from the previous tests that were written so I choose to follow the style of the code already present. Here Given is an initialization phase, When is the call function phase, and Then is the check phase of the test from my understanding

PR #372, opened 2024-05-22
Meeting spot notification

mtv47:
Now if you long click on a stop in the Map, a meeting notification is send to all users present in a Trip

Also fixed a bug where already finished suggestions were still being shown on the Map and Stops where hidden because of this.

Added tests for the Notification part of the code as well

### DIFF
--- a/app/src/test/java/com/github/se/wanderpals/service/NotificationsManagerTest.kt
+++ b/app/src/test/java/com/github/se/wanderpals/service/NotificationsManagerTest.kt
@@ -1,0 +1,202 @@
+package com.github.se.wanderpals.service
+
+import com.github.se.wanderpals.model.data.Expense
+import com.github.se.wanderpals.model.data.Stop
+import com.github.se.wanderpals.model.data.TripNotification
+import com.github.se.wanderpals.model.repository.TripsRepository
+import com.github.se.wanderpals.navigationActions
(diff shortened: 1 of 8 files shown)


tosigame (changes requested):
Looks good, clean way to add an announcement. However the code coverage is low, and you should test the NotificationManager method you added. (Its easy, you just have to mockk the repository and you can see how its done in the viewmodel tests). Hope it helps!, good luck!

mtv47:
> Looks good, clean way to add an announcement. However the code coverage is low, and you should test the NotificationManager method you added. (Its easy, you just have to mockk the repository and you can see how its done in the viewmodel tests). Hope it helps!, good luck!

I've added a NotificationsManagerTest to test all methods but unfortunately I wasn't able to mock LocalDateTime.now()

app/src/test/java/com/github/se/wanderpals/service/NotificationsManagerTest.kt
tosigame, line 1: Great, you tested everything in the new function. And even tested the whole NotificationManager, while at it. Great job!

tosigame (approved):
Great job, works fine for me , and good job on testing the whole class, and thank you for your responsiveness.  As always excellent Job 🎖️

PR #373, opened 2024-05-22
Currency management

Exino329:
In this PR, currency management have been done :
- If not a viewer, a member of a trip can change the Currency used for finance in the trip.
- He can search the currency by either searching its name or its currency code (case insensitive)
- On change, the currency symbol is updated everywhere in the Finance screen. It is also updated on the dashboard.

(Note : there is no currency conversion for the amount, it is just a visual change to better reflect the currency that will be used for expenses in the trip)

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
@@ -41,7 +41,7 @@ open class DashboardViewModel(
   private var _lastPrivateDocument = MutableStateFlow("")
   open val lastPrivateDocument: StateFlow<String> = _lastPrivateDocument.asStateFlow()
 
-  private var _currencyCode= MutableStateFlow("")
+  private var _currencyCode = MutableStateFlow("")
   open val currencyCode: StateFlow<String> = _currencyCode.asStateFlow()
 
   /** Fetches all trips from the repository and updates the state flow accordingly. */
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
@@ -42,7 +42,7 @@ open class DashboardViewModel(
   open val lastPrivateDocument: StateFlow<String> = _lastPrivateDocument.asStateFlow()
 
   private var _currencyCode= MutableStateFlow("")
-  val currencyCode: StateFlow<String> = _currencyCode.asStateFlow()
+  open val currencyCode: StateFlow<String> = _currencyCode.asStateFlow()
 
   /** Fetches all trips from the repository and updates the state flow accordingly. */
   open fun loadSuggestion(tripId: String) {
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
@@ -41,6 +41,9 @@ open class DashboardViewModel(
   private var _lastPrivateDocument = MutableStateFlow("")
   open val lastPrivateDocument: StateFlow<String> = _lastPrivateDocument.asStateFlow()
 
+  private var _currencyCode= MutableStateFlow("")
+  val currencyCode: StateFlow<String> = _currencyCode.asStateFlow()
+
   /** Fetches all trips from the repository and updates the state flow accordingly. */
   open fun loadSuggestion(tripId: String) {
     viewModelScope.launch {
@@ -55,6 +58,7 @@ open class DashboardViewModel(
     viewModelScope.launch {
       // Fetch all expenses from the trip
       _expenses.value = tripsRepository.getAllExpensesFromTrip(tripId)
+      _currencyCode.value = tripsRepository.getTrip(tripId)!!.currencyCode
     }
   }
 
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
@@ -57,11 +57,10 @@ fun DashboardFinanceWidget(viewModel: DashboardViewModel, onClick: () -> Unit = 
 
   ElevatedCard(
       modifier =
-      Modifier
-          .padding(horizontal = 16.dp)
-          .fillMaxWidth()
-          .clickable(onClick = onClick)
-          .testTag("financeCard"),
+          Modifier.padding(horizontal = 16.dp)
+              .fillMaxWidth()
+              .clickable(onClick = onClick)
+              .testTag("financeCard"),
       colors =
           CardDefaults.cardColors(
               containerColor =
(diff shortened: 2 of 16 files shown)


app/src/main/java/com/github/se/wanderpals/model/viewmodel/DashboardViewModel.kt
```diff
@@ -41,6 +41,9 @@ open class DashboardViewModel(
   private var _lastPrivateDocument = MutableStateFlow("")
   open val lastPrivateDocument: StateFlow<String> = _lastPrivateDocument.asStateFlow()
 
+  private var _currencyCode = MutableStateFlow("")
```
mtv47, line 44: You can set it to the default CHF as you did for Trip

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
```diff
@@ -46,11 +47,14 @@ import com.github.se.wanderpals.ui.screens.trip.finance.FinancePieChart
  * @param viewModel The ViewModel for the Dashboard screen.
  * @param onClick The callback function for when the widget is clicked.
  */
+@SuppressLint("DefaultLocale")
```
mtv47, line 50: When you use String.format without explicitly specifying a locale, it defaults to the system’s default locale. However, this can lead to unexpected behavior in certain situations. Instead use for example: String.format(Locale.US,...)

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/FinancePieChart.kt
```diff
@@ -26,13 +26,15 @@ import com.github.se.wanderpals.model.data.Expense
...
+ * @param currencySymbol Symbol of the currency.
  */
 @Composable
 fun FinancePieChart(
     expenses: List<Expense>,
     radiusOuter: Dp = 100.dp,
     chartBandWidth: Dp = 20.dp,
-    totalValueDisplayIsEnabled: Boolean = false
+    totalValueDisplayIsEnabled: Boolean = false,
+    currencySymbol: String = "CHF"
```
mtv47, line 37: You should remove the default setting here to stay consistent with the rest of the code since for other arguments in other functions, you didn't set it. Or make it a variable in strings.xml

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/FinanceTopBar.kt
```diff
@@ -39,35 +43,63 @@ import com.github.se.wanderpals.ui.navigation.Route
...
-                  style =
-                      TextStyle(
-                          fontSize = 24.sp, // Taille de police personnalisée
-                          fontWeight = FontWeight.Bold // Police en gras
-                          ),
-                  modifier = Modifier.padding(start = 30.dp))
+                  modifier = Modifier.padding(end = 20.dp).width(80.dp).testTag("currencyButton"),
+                  onClick = { onCurrencyClick() },
+                  enabled =
+                      SessionManager.getIsNetworkAvailable() &&
```
mtv47, line 89: Nice to have taken into account the offline/online feature

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/FinanceTopBar.kt
```diff
@@ -39,35 +43,63 @@ import com.github.se.wanderpals.ui.navigation.Route
...
+                          modifier = Modifier.size(35.dp),
+                          imageVector = Icons.AutoMirrored.Rounded.ArrowBack,
+                          contentDescription = "Back",
+                      )
+                    }
+                    Text(
+                        modifier = Modifier.padding(start = 20.dp),
+                        text = "Finance",
+                        textAlign = TextAlign.Center,
+                        style = TextStyle(fontSize = 24.sp, fontWeight = FontWeight.Bold),
```
mtv47, line 81: You should stay consistent with your use of MaterialTheme

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @param financeViewModel The [FinanceViewModel] used to manage the currency selection state.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun CurrencySelectionDialog(financeViewModel: FinanceViewModel) {
+
+  var searchedCurrency by remember { mutableStateOf("") }
+  var isError by remember { mutableStateOf(false) }
+  val currencies = Currency.getAvailableCurrencies().filterNot { it.displayName.contains("(") }
```
mtv47, line 56: Weird patch but works nicely, concise and not complex

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+                    value = searchedCurrency,
+                    onValueChange = { value -> searchedCurrency = value },
+                    modifier =
+                        Modifier.fillMaxWidth()
+                            .padding(20.dp)
+                            .clip(RoundedCornerShape(5.dp))
+                            .testTag("currencySearchText"),
+                    label = {
+                      Text(
+                          text = if (isError) "Invalid currency" else "Select a currency",
```
mtv47, line 85: Could have used when then were instead of if else to reduce the cognitive complexity. From my understanding, when then doesn't count but if else does

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+                                        RoundedCornerShape(5.dp),
+                                    )
+                                    .testTag("currencyItem"),
+                            contentAlignment = Alignment.CenterStart) {
+                              Button(
+                                  modifier = Modifier.fillMaxWidth(),
+                                  colors =
+                                      ButtonDefaults.buttonColors(
+                                          containerColor = Color.Transparent),
+                                  onClick = { searchedCurrency = currency.displayName }) {}
```
mtv47, line 154: Could remove the {}

mtv47 (changes requested):
@SuppressLint("DefaultLocale") should be removed since it's an easy fix and can cause problems. Otherwise very clean code expect for the small inconsistency in the setting or not the currency by default

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
Exino329: Thank you for catching this, I didn't know this.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
Exino329: Here using when seems a bit overkilled since we have just an if else case

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
Exino329: I have to do it because the button has no content but I need to specify it to make the app compile

mtv47 (approved):
👍

PR #375, opened 2024-05-23
Settle debt

AgeX21:

- It's possible to settle the debt of users who owe money to the current in the debt screen, it creates an expense with the right name


ForestDope (changes requested):
Overall good but some small changes are necessary and you should try to reduce the cognitive complexity of DebtContent

ForestDope (approved):
Good work.

PR #376, opened 2024-05-23
Fix UI bug display in Overview

Exino329:
The goal of this PR was to fix 2 UI bugs in the overview :
- If the trip title was too long, it was moving the start and end date outside of the overview trip Card and was hiding them.
- When clicking on send,share or document icon button of the overview, a small visual effect was triggered. However when clicking on the share icon button, the effect was also triggered for the document button.

All of theses bugs have been fixed in this PR. Also a test has been added to check that trip title is handled correctly when it is too long (i.e. when it overflows)


ForestDope (approved):
LGTM

Issue #361, opened 2024-05-19
Develop the agenda markers for more status associated with different colors.


Polo150202:
1) Develop the agenda markers for more status associated with different colors.

PR #377, opened 2024-05-23
Agenda markers; add calendar trip range background; refactor markers related part of the code

annhit29:
In this pull request,
1) 
I have developed the agenda markers to have more statuses associated with different colors on the calendar of the agenda.
There are four statuses for the StopStatus enumeration class:
1) the `CURRENT` status which stands for a suggestion that has been transformed to a stop of the trip and it is a stop of the trip of the current day;
2) the `COMING_SOON` status which stands for a suggestion that has been transformed to a stop of the trip and it is a stop of the trip on a day between the next day of the current day and the end of the trip date (included);
3) the `PAST` status which stands for a suggestion that has been transformed to a stop of the trip and it is a stop of the trip on a day between the the start of the trip date (included) and the day before the current day (included);
4) the `NONE` status which stands for a suggestion that has not been transformed to a stop of the trip yet, it is still a suggestion of the trip.

The colors of the stop statuses are dynamic. Meaning:
- If the current day is 2024 May 20th and there is a stop on that day, then the marker color will be purple on that day, and for the other days of the trip: the marker color will be either grey if the days where there is a stop is before 2024 May 20th or light blue if the days where there is a stop is after 2024 May 20th.
- If we come back to the app the next day (for example), the marker color for the day 2024 May 20th will be grey, and if there are stops on 2024 May 21st, then the marker color for 2024 May 21st will be purple, and for the other days of the trip: the marker color will be either grey if the days where there is a stop is before 2024 May 21st or light blue if the days where there is a stop is after 2024 May 21st.
If there are no stops on 2024 May 21st, then there will be no markers on that day on the calendar.

\
Note that the `ADDED` status has been removed and there are `CURRENT`, `PAST`, and `COMING_SOON`.

\
2) (after discussing with TAs, this pull request will be light if I only add more `stopStatus`es because it is a functionality that I have already developed, thus here is a new functionality:) There is now a continuous light-purple background to indicate the trip range. For example, if the trip starts on May 15th, 2024, and ends on May 24th, 2024, then the background within this trip range will be displayed on the calendar.

\
3) I have refactored the code related to 1) and 2).

\
\
Therefore, more tests are built in AgendaTest.kt and AgendaViewModelTest.kt for testing the UI elements and the functionalities of 1) and 2).
\
This pull request is for issue #361 .

### DIFF
--- a/app/src/androidTest/java/com/github/se/wanderpals/agenda/AgendaTest.kt
+++ b/app/src/androidTest/java/com/github/se/wanderpals/agenda/AgendaTest.kt
@@ -10,6 +10,7 @@ import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.wanderpals.model.data.Stop
+import com.github.se.wanderpals.model.data.Trip
 import com.github.se.wanderpals.model.repository.TripsRepository
 import com.github.se.wanderpals.model.viewmodel.AgendaViewModel
 import com.github.se.wanderpals.ui.screens.trip.agenda.Agenda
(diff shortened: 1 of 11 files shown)


Polo150202 (approved):
Nice code and changed, maybe we can change the shape, when we fly over the dates, to be circular and not square. But good code

annhit29:
Hi Paul, to answer your concern, the shapes are squares rather than circular, because
if when we fly over the dates with the circular shape, it will give us the following calendar:
image
and it is visually unpleasant.
Thus, I changed to
```kt
  val backgroundShape =
      when {
        isStartDate || isStartOfLine -> RoundedCornerShape(topStart = 50.dp, bottomStart = 50.dp)
        isEndDate || isEndOfLine -> RoundedCornerShape(topEnd = 50.dp, bottomEnd = 50.dp)
        else -> RoundedCornerShape(0.dp)
      }
```
where `RoundedCornerShape(0.dp)` is a square only if the date displayed on the calendar is neither the `startDate` nor the `endDate` of the trip, **and** it is not displayed at the beginning and not at the end of a row of the calendar.

With the code, we get:
image

PR #378, opened 2024-05-23
Fix user name displayed in app 

ForestDope:
In this PR : 

- Upon changing name in the admin panel all places where the name was previously displayed are updated to display the new name.
- The exception is notification has the notification item is not associated with a user directly. 

A first branch trying to do this differently was made but in the end this way is much more clear and works perfectly.


AgeX21 (approved):
LGTM !

PR #380, opened 2024-05-24
Redesign a menu for the dashboard

AgeX21:

- Changed the menu to be more in theme with the rest of the app and serve as a central point to navigating around
- Fixed a display issue in the widget of suggestion and stop where a horizontal divider would display where it shouldn't
- Added a share button in the dashboard to complete every feature the dashboard should have based on the figma

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardStopWidget.kt
+++ b/app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardStopWidget.kt
@@ -138,11 +138,11 @@ fun DashboardStopWidget(viewModel: DashboardViewModel, onClick: () -> Unit = {})
                       } else {
                         Column {
                           StopWidgetItem(stop = sortedStops[0])
-                          HorizontalDivider(
-                              color = MaterialTheme.colorScheme.surfaceVariant,
-                              thickness = 1.dp,
-                              modifier = Modifier.padding(horizontal = 8.dp))
                           if (sortedStops.size > 1) {
+                            HorizontalDivider(
+                                color = MaterialTheme.colorScheme.surfaceVariant,
+                                thickness = 1.dp,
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardStopWidget.kt
```diff
@@ -138,11 +138,11 @@ fun DashboardStopWidget(viewModel: DashboardViewModel, onClick: () -> Unit = {})
                       } else {
                         Column {
                           StopWidgetItem(stop = sortedStops[0])
-                          HorizontalDivider(
-                              color = MaterialTheme.colorScheme.surfaceVariant,
-                              thickness = 1.dp,
-                              modifier = Modifier.padding(horizontal = 8.dp))
                           if (sortedStops.size > 1) {
+                            HorizontalDivider(
```
utkarsh006, line 142: Can create a separate function for divider component which will receive all the details in its parameters, and it can be called anywhere in the code.

ForestDope: Thank you for your interest in our project. However for the moment this repository is actively being graded for our course.

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/Dashboard.kt
ForestDope, line 1: Cool use of val for the colors, makes the code really modular :D

ForestDope (changes requested):
Overall very good code, you could just reduce the size of the share button slightly and also write a test for it.

AgeX21:
I thought about it and my decision was that if this menu is added to other screen, dashboard should then appear so maybe we should remove the current screen from the list... But I think we settled down on having only the dashboard display a menu.

ForestDope (approved):
LGTM

PR #381, opened 2024-05-24
Server to secure and avoid depreciation

Polo150202:


app/src/main/java/com/github/se/wanderpals/MainActivity.kt
```diff
@@ -317,6 +312,9 @@ class MainActivity : ComponentActivity() {
                   Trip(navigationActions, tripId, viewModel.getTripsRepository(), mapManager)
                 }
                 composable(Route.CREATE_TRIP) {
+                  NotificationAPI()
+                      .sendNotification(listOf(SessionManager.getNotificationToken()), "hello2")
```
annhit29, line 316: You could replace `hello2` with another string with a more meaningful `body` name, since the function `sendNotification` of NotificationAPI.kt needs one `body: String`.

annhit29 (changes requested):
In general, very good! Please
1) add block comment on the BroadcastNotification data class;
2) replace `hello2` with a more meaningful body name;
3) Create or modify tests for the function `addAnnouncement` of NotificationsViewModel;
4) Add a description of what your pull request does.

app/src/main/java/com/github/se/wanderpals/model/data/BroadcastNotification.kt
annhit29, line 1: Great!

app/src/main/java/com/github/se/wanderpals/model/viewmodel/NotificationAPI.kt
annhit29, line 1: Good job!

annhit29 (approved):
This looks good to me! Please just add a description of what your pull request does.

The line coverage is above 80% which is great too :D

Issue #386, opened 2024-05-25
Currency automatic conversion


ForestDope:
The user should be able to not only see expenses in different currencies (already the case) but also the currency should be automatically converted to reflect the correct amount.

PR #389, opened 2024-05-25
Currency conversion in real time 

Exino329:
This PR aimed to perform real time currency when the user changes the currency of the trip (this PR follows the work did on PR #386).
-  Now when changing the currency, the FinanceViewModel will perform an http get request using `OkHttp` on the currency converter coming from this website : [](https://www.oanda.com/rw-en/trading/) and use the old currency and new currency in the queries parameters
-  The response is parsed as a json answer response from their currency converter. After extracting the json answer containing the exchange rate between both currencies, the financeViewModel performs the expense amount transformation accordingly.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
@@ -1,9 +1,9 @@
 package com.github.se.wanderpals.model.viewmodel
 
+import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
-import androidx.lifecycle.viewmodel.compose.viewModel
 import com.github.se.wanderpals.model.data.Expense
 import com.github.se.wanderpals.model.data.User
 import com.github.se.wanderpals.model.repository.TripsRepository
@@ -19,7 +19,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
-import okhttp3.Dispatcher
+import okhttp3.HttpUrl
 import org.json.JSONObject
 import java.io.IOException
 import java.time.LocalDate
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/FinanceBottomBar.kt
```diff
@@ -26,6 +26,8 @@ import com.github.se.wanderpals.service.SessionManager
  */
 @Composable
 fun FinanceBottomBar(expenses: List<Expense>, currencySymbol: String) {
+  val totalAmount =
```
mtv47, line 29: Great choice of moving it outside of text

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
mtv47, line 1: I don't see why it has changed, I think you should go back to the previous version of this file and avoid creating a variable for no reason

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
```diff
@@ -96,19 +108,101 @@ open class FinanceViewModel(val tripsRepository: TripsRepository, val tripId: St
...
+   * @param newCurrencyCode The new currency code to be used for the trip.
    */
-  open fun updateCurrency(currencyCode: String) {
+  open fun updateCurrency(newCurrencyCode: String) {
     viewModelScope.launch {
       val currentTrip = tripsRepository.getTrip(tripId)!!
-      tripsRepository.updateTrip(currentTrip.copy(currencyCode = currencyCode))
-      updateStateLists()
+      updateExchangeRate(currentTrip.currencyCode, newCurrencyCode)
+      if (exchangeRate.value != null) {
```
mtv47, line 121: You can put this if block inside of updateExchangeRate to not check again if it was successfull, since it's called "update". Or call it set or get

mtv47 (dismissed):
Overall useful and clean code, some changes can be made but are not required

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CurrencySelectionDialog.kt
Exino329: Effectively, I initally made some changes in the methods of FinanceViewModel and finally reverted them, and I forgot to re-apply those changes in CurrencySelectionDialog thank you for catching it !

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
Exino329: Yes it makes more sens I will change the name of the method.

app/src/test/java/com/github/se/wanderpals/viewmodel/FinanceViewModelTest.kt
```diff
@@ -93,12 +115,43 @@ class FinanceViewModelTest {
...
+        // Arrange
+        val newCurrency = "EUR"
+        val exchangeRate = 0.85
+
+        // Mocking the _exchangeRate LiveData
+        val exchangeRateLiveData = MutableStateFlow<Double?>(exchangeRate)
+        viewModel.exchangeRate.value = exchangeRateLiveData.value
+
+        val viewModelSpy = spyk(viewModel, recordPrivateCalls = true)
+        viewModelSpy
```
mtv47, line 140: You can delete viewModelSpy here

mtv47 (approved):
All good

PR #391, opened 2024-05-27
Fix Expense Duplication and Screen Flicker in FinanceViewModel and CreateExpense Screen

tosigame:
This PR addresses two critical issues in the FinanceViewModel and CreateExpense screen. Firstly, it resolves the expense duplication bug by ensuring addExpense is not called multiple times when invoked rapidly. Secondly, it fixes the screen flicker issue by properly loading data before displaying the expense list, preventing the initial empty state. Unit tests have been added to verify the bug.

### DIFF
--- a/app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
+++ b/app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
@@ -34,7 +34,7 @@ import org.json.JSONObject
  * @param tripId The ID of the trip this ViewModel is associated with.
  */
 open class FinanceViewModel(val tripsRepository: TripsRepository, val tripId: String) :
-    ViewModel() {
+  ViewModel() {
 
   private val _users = MutableStateFlow(emptyList<User>())
   open val users: StateFlow<List<User>> = _users.asStateFlow()
@@ -56,6 +56,15 @@ open class FinanceViewModel(val tripsRepository: TripsRepository, val tripId: St
 
   private val _tripCurrency = MutableStateFlow<Currency>(Currency.getInstance("CHF"))
   open val tripCurrency = _tripCurrency.asStateFlow()
+
+  // signal that the Expense was added successfully
+  private val _createExpenseFinished = MutableStateFlow(false)
+  open val createExpenseFinished: StateFlow<Boolean> =
+    _createExpenseFinished.asStateFlow()
+
+  // don't add an Expense twice
+  private val _isAddingExpense = MutableStateFlow(false)
+  open val isAddingExpense: StateFlow<Boolean> = _isAddingExpense.asStateFlow()
 
   val exchangeRate = MutableStateFlow<Double?>(null)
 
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/github/se/wanderpals/viewmodel/FinanceViewModelTest.kt
mtv47, line 1: The test is very thought out and proves that the fix is really working. Great job in putting in the effort.

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
```diff
@@ -83,14 +91,31 @@ open class FinanceViewModel(val tripsRepository: TripsRepository, val tripId: St
...
+        if (expenseList.isEmpty()) {
+          // error
+          Log.e("FinanceViewModel", "Error reading expense after addition (should not happen)")
+        } else {
+          val newExpense = expenseList.last()
+          NotificationsManager.addExpenseNotification(tripId, newExpense)
+        }
+        updateStateLists()
+        _isAddingExpense.value = false
+        _createExpenseFinished.value = true
```
mtv47, line 109: Could have used setCreateExpenseFinished for more readability

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
```diff
@@ -83,14 +91,31 @@ open class FinanceViewModel(val tripsRepository: TripsRepository, val tripId: St
...
+
     viewModelScope.launch {
-      val newExpense = tripsRepository.getAllExpensesFromTrip(tripId).last()
-      NotificationsManager.addExpenseNotification(tripId, newExpense)
-      updateStateLists()
+      if (!isAddingExpense.value && !createExpenseFinished.value) {
+        _isAddingExpense.value = true
+        tripsRepository.addExpenseToTrip(tripId, expense)
+        val expenseList = tripsRepository.getAllExpensesFromTrip(tripId)
+        if (expenseList.isEmpty()) {
```
mtv47, line 100: Why not, more error checking is always good. Isn't relevant to this PR but is appreciated

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/CreateExpense.kt
mtv47, line 1: Clean

mtv47 (approved):
Efficient and nice PR. Fixes a bug that is important with additional written tests that check that the bug won't appear further down the line

app/src/main/java/com/github/se/wanderpals/model/viewmodel/FinanceViewModel.kt
tosigame: You may be right, however my idea behind setCreateExpenseFinished, was for the UI to signal to the viewModel, that it has finished getting the information from the update.

PR #395, opened 2024-05-28
Redesign finance topbar

AgeX21:
in this PR, the following has been done : 
- changed the topbar of finance to share the same graphical identity as the other screens


app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/ExpensesContent.kt
annhit29, line 1: This looks nice!

app/src/main/java/com/github/se/wanderpals/ui/screens/trip/finance/FinanceTopBar.kt
annhit29, line 1: This makes every screen consistent, i.e. having the same layout and styling for the topbar, nice!

annhit29 (approved):
Good job for making the Finance top bar nice-looking and consistent with the topbars of other screens!

PR #397, opened 2024-05-28
Redesign Debt screen

AgeX21:

- Changed the color of the card and text to make it more pretty and in theme with the app : balance is green or red depending on the global debt a user has.


annhit29 (approved):
This looks good to me :D
Well done!

PR #398, opened 2024-05-29
Fix bug of the pie chart display

Polo150202:


app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
```diff
@@ -212,3 +213,39 @@ fun ExpenseItem(expense: Expense, currencyCode: String) {
...
+          2 -> (expenseToString.reversed() + "M")
+          3 -> (expenseToString.reversed() + "B")
+          4 -> (expenseToString.reversed() + "T")
+          else -> {
+            "Error"
+          }
+        }
+  } else {
+    formattedString = formattedString.reversed()
+    Log.d("formattedString", formattedString)
```
tosigame, line 247: The Log.d statement in the formatExpense function is useful for debugging. Should it be left in the final version?

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
```diff
@@ -212,3 +213,39 @@ fun ExpenseItem(expense: Expense, currencyCode: String) {
...
+
+  val numString = formattedString.substring(3)
+  val numberOfDigits = numString.length / 3
+  if (numberOfDigits > 1) {
+    val formattedReversedString = numString.chunked(3).joinToString(".")
+
+    // take only the digits after the last point and the two digits before the point
+    val expenseToString = formattedReversedString.substringAfterLast(".").substringBeforeLast(".")
+
+    formattedString =
```
tosigame, line 236: Have you tested the formatExpense function with various edge cases? For example, very small or very large values? It would be good to ensure it handles all scenarios gracefully.

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
```diff
@@ -212,3 +213,39 @@ fun ExpenseItem(expense: Expense, currencyCode: String) {
...
+
+/**
+ * Function to format the total expense amount to a more readable format. The function takes the
+ * total expense amount as a Double and returns a formatted String. The function formats the amount
+ * to a maximum of 2 decimal places and adds a suffix of M, B, or T based on the number of digits in
+ * the amount.
+ *
+ * @param totalexpense The total expense amount as a Double.
+ */
+private fun formatExpense(totalexpense: Double): String {
```
tosigame, line 225: Consider moving this function out of the screen and into the ViewModel for better encapsulation and easier testing with unit tests.

tosigame (changes requested):
Overall, this PR effectively addresses the issue and introduces a useful new utility function. However, I suggest adding tests for this function. Additionally, consider moving the function into the ViewModel. This change would clean up the screen code and facilitate testing by allowing for unit tests.

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
Polo150202: I will add some tests to see

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardFinanceWidget.kt
Polo150202: Sure, i have to remove it

app/src/androidTest/java/com/github/se/wanderpals/dashboard/DashboardTest.kt
tosigame, line 1: Great tests congrats on the 100% coverage

tosigame (approved):
Great work on the tests! It seems you found the Android tests to be more coherent for the situation. I partially agree with this because your function is only used in a single place, which avoids polluting the ViewModel that may be reused elsewhere.

PR #399, opened 2024-05-29
Fixed a bug in Suggestion Widget

AgeX21:

- Fixed a bug in the suggestion widget where accepted suggestion would still appear on the dashboard
- Added a regression test to prevent this bug from reappearing


app/src/androidTest/java/com/github/se/wanderpals/dashboard/DashboardTest.kt
tosigame, line 1: I appreciate the addition of the regression test case to ensure that deleted suggestions are not displayed. This will help maintain the integrity of the suggestions feature.

app/src/main/java/com/github/se/wanderpals/ui/screens/dashboard/DashboardSuggestionWidget.kt
tosigame, line 1: Simple and efficient :+1:

tosigame (approved):
The changes look great! The new test case and modifications to the suggestions widget are well-implemented. Thanks for addressing the issues. :medal_military:

PR #403, opened 2024-05-30
Redesign Document Widget

AgeX21:

- Redesigned the Document Widget to fit more with the theme of the App
- Added test for the widget
- Fixed a small graphical issue where a white bar would be visible at the bottom of the screen if the screen was full


annhit29 (approved):
Thank you for redesigning the document widget! It is consistent with every widget of the dashboard component!

The redesigned document widget:
image

The entire screen of Dashboard:
image
As we see from the screenshot,
> Fixed a small graphical issue where a white bar would be visible at the bottom of the screen if the screen was full
this is fixed.

Thank you for adding the test in DashboardTest.kt, congratulations on the 100% in line coverage!

PR #406, opened 2024-05-30
Update Architecture diagramm and added text to README

tosigame:
Update Architecture diagramm and added text to README


Polo150202 (approved):
Great Diagram and description

PR #407, opened 2024-05-31
Fix the displaying of the Trip's Image

Polo150202:
I added the Async Image compose to be able to display the Image that we stored in Cloud Storage of Firebase.


AgeX21 (approved):
LGTM !

PR #408, opened 2024-05-31
Minor bug found and fix

Polo150202:
Just a Minor Bug found, that unable us to accept and add the document due to a false condition


AgeX21 (approved):
LGTM !

PR #412, opened 2024-06-01
Remove secret inside the code to limit security issue

Polo150202:
Remove secret inside the code to limit security issue inside the String.xml


AgeX21 (approved):
Good idea !

